PR #1, opened 2018-01-22
Fix manifest for @PlatformIO

ivankravets:
How about to rename `esphomelib` to `src`? this is default structure for libraries.

### DIFF
--- a/library.json
+++ b/library.json
@@ -36,7 +36,8 @@
       "version": "~3.7.8"
     },
     {
-      "name": "https://github.com/OttoWinter/PCA9685-Arduino.git"
+      "name": "PCA9685-Arduino",
+      "version": "https://github.com/OttoWinter/PCA9685-Arduino.git"
     }
   ],
   "build": {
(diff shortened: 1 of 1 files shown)


OttoWinter:
Thanks! The reason having all the headers (and source files) under the `esphomelib` folder was that all `#include`s would be of the form `<esphomelib/...>`, but I guess putting the current `esphomelib` folder into a new `src` folder could accomplish that too.

Another thing I just noticed: The platformio library registry doesn't currently show any examples, even though the `"examples": "examples/*.cpp"` looks fine to me. Do I have to change something?

ivankravets:
You can use `src` and `include` project structure, where you can keep header files in `include` folder. PIO will add `include` folder to CPPPTH automatically.

> Do I have to change something?

I don't remember... :( It seems when examples are outside of library root, PIO does not see them. In your manifest, you ask PIO to use `esphomelib` as a ROOT of library. See docs http://docs.platformio.org/en/latest/librarymanager/config.html#include

You can remove `export` section from `library.json` and configure build process via [build](http://docs.platformio.org/en/latest/librarymanager/config.html#build) section.

PR #4, opened 2018-01-23
MQTT topic_prefix and sanitization

OttoWinter:
This PR allows users to specify a topic_prefix for MQTT, so that prefixes like `home/livingroom/...` can be used. Additionally, lots of sanitization is now done throughout the library for hostnames, discovery, etc. and the documentation is updated to reflect the changes.


OttoWinter:
Tested locally and it works, merging this.

PR #7, opened 2018-01-30
Add Home Assistant configuration sample

fabaff:
Add Home Assistant configuration sample and some other minor changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ void loop() {
 }
 ```
 
-And voilà - esphomelib will now automatically manage everything such as logging to MQTT, Home Assistant MQTT-discovery, 
+And voilà - esphomelib will now automatically manage everything such as logging to MQTT, Home Assistant MQTT discovery, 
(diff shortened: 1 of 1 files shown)


OttoWinter:
Thanks 🌻

OttoWinter:
Oh, I didn't realize the default Home Assistant MQTT discovery prefix was `homeassistant/`. On second thought, as discovery will only ever be used in with Home Assistant it might be more user friendly to just use Home Assistant's default. But on the other hand, `discovery/` makes more sense to me and users have to do explicitly enable MQTT discovery anyway.

fabaff:
It's doesn't really matter how the topic is called, from my point of view. When I wrote the MQTT discovery part of Home Assistant I want to go with a wildcard but Paulus had the idea of allowing/using a prefix additionally to better distinguish between the topics and make it possible for other solutions to re-use the same technique.

The idealistic idea of being able to setup something and use it with what ever home automation solution or tool you want without any changes should be the main goal of all IoT development.

PR #8, opened 2018-02-18
ESP8266 support

OttoWinter:
TODO:

- [x] Documentation
- [x] Some more testing
- [x] Fix travis
- [x] Fix library.json
- [ ] Preferences
- [ ] IR without esp-idf remote (should be relatively simple)


madrover:
Very looking forward to this one! :)

OttoWinter:
@madrover Me too 😬. I've been spending lots of time with the Home Assistant project for the last few weeks but I plan on adding fresh new functionality to this project soon.

It's just that for every component I test here I first need to create the hardware for it first in order to be able to test for it, and that takes some time with my far from perfect soldering skillz 😅.

madrover:
I want to implement [ColorWavesWithPalettes.ino](https://gist.github.com/kriegsman/8281905786e8b2632aeb) with an ESP8266 being controlled by HA. Your framework looks like an interesting foundation for this project, so I don't mind to betatest it.
The hardware aspect of the project is pretty settled and I now need to focus on the software side.
Any recommendation on how to approach this?
Shall I just checkout the esp8266-support branch, start coding and report issues? I can try to fix some stuff, but my skills in C++ are rather limited...

OttoWinter:
Supporting FastLED devices is something I really want to support, the problem is: they offer **so much** customizability that creating any API for them in esphomelib would just be madness and eventually everyone would have to override the API themselves to achieve their own effects. That and my lack of FastLED lights have been what has held me from actually implementing them.

esphomelib is great at handling the communication with Home Assistant for you. I think the best way for you to have a try at supporting FastLED for your own setup would be to create your own `LightOutput`:

https://github.com/OttoWinter/esphomelib/blob/95ff3365bbedf28cbe9be60201f16c3d383b700e/src/esphomelib/light/light_output_component.h#L18-L31

If you override this class, your new `LightOutput` will be repeatedly called with the **state** of your lights. In this state you can store stuff like the RGB values, the active effective and so on (see `set_state`).

The next step would then be to actually connect your `LightOutput` with your MQTT Light, so that the MQTT Light can tell your output what values are currently active. I'm in the process of writing docs for how to do this, but see how the Application instance does this for some reference:

https://github.com/OttoWinter/esphomelib/blob/95ff3365bbedf28cbe9be60201f16c3d383b700e/src/esphomelib/application.cpp#L223-L232

After that's hopefully working, you can start with implementing effects for your new Light. For this, just have a look at the [LightEffect class](https://github.com/OttoWinter/esphomelib/blob/master/src/esphomelib/light/light_effect.h) and just follow the RandomLightEffect, for example. Register your effect with `light_effect_entries.push_back(**your_effect_constructor**)` and that should more or less be it.

If you have questions, please ask :)

PR #9, opened 2018-02-18
Documentation

OttoWinter:
TODO:

- [ ] Getting started guides
  - [ ] Light
  - [ ] GPIO
  - [ ] Custom Components
  - [ ] Internals
  - [ ] ESP8266
  - [ ] Usage without Home Asistant
- [ ] API reference
  - [ ] Complete in-code documentation
  - [ ] Binary Sensor
  - [ ] Sensor
  - [ ] Light
  - [ ] Fan
  - [ ] Input
  - [ ] Output
- [ ] Create readthedocs website
- [ ] About, Contributing, Copyright


Evanlec (approved):
Looks good to me. For documentation, I'd like to see how I could get the values out of a sensor, for output to an OLED display for example. I am not well versed enough in C++ apparently, since I could not figure out what method to call / override by reading your code.

Really like this library though. Handles all the wifi and mqtt stuff for me which makes the code much cleaner.

OttoWinter:
@Evanlec Yes I'm writing a guide for creating sensors as we speak 😀. May I ask you what exactly you want to do with your OLED display, because I think with the current API it might be a bit difficult (but that should be easy to fix, PR incoming 😬)

(The things mentioned in the guides will not be final though, and the API might change; I don't want to restrict myself to the current API for now, as the library is still quite new; but https://github.com/OttoWinter/esphomelib/pull/20 should make sensors a *lot* easier to implement)

Also, I probably will scrap this PR because I've now kind of decided that I want to host the guides on the GitHub wiki and just host the auto-generated API on readthedocs.

OttoWinter:
@Evanlec An example for creating custom sensors is now up at https://github.com/OttoWinter/esphomelib/wiki/Custom-Sensor-Component. I plan on doing the same for output components next.

Evanlec:
Ah excellent, thank @OttoWinter  -- as for the OLED -- I simply want to display the temperature/humidity from my sensor.  And for my thermostat relay, perhaps display the target temperature setting (and maybe allow buttons to increase/decrease target temp setting).

OttoWinter:
Closing for #24. A documentation is now available at http://esphomelib.com

Issue #26, opened 2018-03-25
Compile error ESP8266 


geelenbert:
Hi,

I'm trying to build something for an NodeMCU v2 device. 

```
#include "esphomelib/application.h"

using namespace esphomelib;

void setup() {
  App.set_name("displates");
  App.init_log();

  App.init_wifi("xxxxxx", "xxxxxxxx");
  App.init_mqtt("192.168.1.xxx", "homeassistant", "xxxxxxxxxxx");
  auto *ota = App.init_ota();
  // Set a plaintext password, alternatively use an MD5 hash for maximum security (set_auth_password_hash)
  ota->set_auth_plaintext_password("knippen5");
  ota->start_safe_mode();

  App.make_monochromatic_light("Displate 1", App.make_ledc_output(13)); // supports brightness
  
  App.setup();
}

void loop() {
  App.loop();
}
```

Compiling the code gives:
```
[Sun Mar 25 19:39:21 2018] Processing displates (platform: espressif8266; board: nodemcuv2; framework: arduin
o)
--------------------------------------------------------------------------------
Verbose mode can be enabled via `-v, --verbose` option
PLATFORM: Espressif 8266 > NodeMCU 1.0 (ESP-12E Module)
SYSTEM: ESP8266 80MHz 80KB RAM (4MB Flash)
Library Dependency Finder -> http://bit.ly/configure-pio-ldf

LDF MODES: FINDER(chain) COMPATIBILITY(light)
Collected 32 compatible libraries
Scanning dependencies...
Dependency Graph
|-- <PubSubClient> v2.6
|-- <ArduinoJson> v5.13.1
|-- <DHT> v1.0.0
|-- <DallasTemperature> v3.8.0
|   |-- <OneWire> v2.3.4
|-- <PCA9685 16-Channel PWM Driver Module Library> v1.2.9 #22b7212
|   |-- <Wire> v1.0
|-- <ESP8266WiFi> v1.0
|-- <ArduinoOTA> v1.0
|   |-- <ESP8266WiFi> v1.0
|   |-- <ESP8266mDNS>
|   |   |-- <ESP8266WiFi> v1.0
Compiling .pioenvs/displates/src/examples/displates.cpp.o
Compiling .pioenvs/displates/src/src/esphomelib/application.cpp.o
Compiling .pioenvs/displates/src/src/esphomelib/binary_sensor/binary_sensor.cpp.o
[...]

PR #27, opened 2018-03-25
ESP8266 Software PWM

OttoWinter:
ESP8266 wasn't too hard to implement and it's probably quite useful even though it's quite limited.

Todo:
  - [x] Some testing

Partial fix for #26


OttoWinter:
Test locally and works, yay 🎉

PR #48, opened 2018-05-18
Make esphomelib compatible with the Arduino IDE

per1234:
- Add library.properties file with library metadata.
- Move library source files directly under the `src` subfolder.
- Move each example program to its own folder and add dummy .ino files to make them valid Arduino sketches.
- Update platformio.ini to reflect the new example program locations.
- Add Arduino IDE installation instructions to README.md.
- Update `#include` directives according to new source files location.
- Make all internal `#include` directives relative.
  - This will avoid searching the full include path for the files to avoid filename collisions with other installed libraries.

I've endeavored to make the fewest number of changes possible. The Arduino IDE user experience could be further enhanced by the following additional changes (which are **not** implemented in this pull request):
- Fix warnings that cause compilation to fail for ESP32 when File > Preferences > Compiler warnings is set to "More" or "All" (Arduino core for the ESP32 [sets `-Werror=all` at those warning levels](https://github.com/espressif/arduino-esp32/blob/master/platform.txt#L20-L21))
- Add the file src/esphomelib.h, which simply #includes application.h and #include esphomelib.h from the examples. This will make it more clear which library is being used and avoid possible name collisions with other installed libraries that have a file named application.h.
- Move example code from the .cpp file to the .ino file. Although I did add an explanatory comment to the dummy .ino files, I think it will be a bit confusing for some Arduino users to find the primary file empty.
- Add esphomelib and to the [Arduino Library Manager index](https://github.com/arduino/Arduino/wiki/Library-Manager-FAQ) for the easiest possible installation and update of the library.

Reference:
https://github.com/arduino/Arduino/wiki/Arduino-IDE-1.5:-Library-specification

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ For example, the software for a device with an RGB light using the internal PWM 
 [DHT22](https://www.adafruit.com/product/385) temperature sensor can be as simple as this:
 
 ```cpp
-#include "esphomelib/application.h"
+#include "esphomelib.h"
 
 using namespace esphomelib;
 
@@ -75,8 +75,7 @@ You can find a documentation for the API of this project at [esphomelib.com](htt
 
 ### Setting Up Development Environment
 
->  Note: Developing in the Arduino IDE is currently not possible, but support for this is being looked into. However, 
-> platformio will remain the preferred method.
+#### PlatformIO
 
 esphomelib is made for use with [platformio](http://platformio.org/), an advanced ecosystem for microcontroller
 development. To get started with coding esphomelib applications, you first have to 
@@ -104,6 +103,25 @@ board = nodemcuv2
 
 Finally, create a new source file in the `src/` folder (for example `main.cpp`) and start coding with esphomelib.
 
+#### Arduino IDE
+
+
+##### Installing the esphomelib library
+
+1. Download the latest release from https://github.com/OttoWinter/esphomelib/releases
+2. (In the Arduino IDE) **Sketch > Include Library > Add .ZIP Library... >** select the downloaded file **> Open**
+
+##### Installing library dependencies
+
+Repeat the above steps with the following libraries:
+- https://github.com/marvinroger/async-mqtt-client/releases
+- https://github.com/OttoWinter/ArduinoJson/releases
+- https://github.com/FastLED/FastLED/releases
+- https://github.com/me-no-dev/ESPAsyncWebServer/archive/master.zip
+- https://github.com/me-no-dev/AsyncTCP/archive/master.zip
+
+After installing esphomelib, you will find a variety of example sketches under **File > Examples > esphomelib**.
+
 ### Bare Bones
 
 Before adding all the desired components to your code, there are a few things you need to set up first.
@@ -111,7 +129,7 @@ Before adding all the desired components to your code, there are a few things yo
 Begin with including the library and setting the C++ namespace.
 
 ```cpp
-#include "esphomelib/application.h"
+#include "esphomelib.h"
 
 using namespace esphomelib;
 ```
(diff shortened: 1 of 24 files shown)


OttoWinter:
Wow! Thanks for doing all this work!

A few notes:
 * I agree the example codes should ideally be converted to `.ino` files. Of course can also be in a future PR if you don't want to include it here.
 * Having an `esphomelib.h` global file would definitely help with Arduino compatability.

Now to the library structure: So I personally do quite like the `src/esphomelib/*` directory structure. My question therefore is: does this old directory structure not work with the Arduino IDE? I've seen some notes on the Internet about arduino only searching subfolders non-recursively, which would of course break esphomelib's sub-sub-folder format.

The library.properties specification this is a bit unclear about this fact and it would seem very weird to me if a semi-modern compiler/build framework wouldn't support this, so I thought I'd rather ask.

Reasons I like the recursive format:
 * It's a lot cleaner. If I type `#include "esphomelib/sensor/dht_something.h"` it's immediately clear that this is from esphomelib as opposed to `#include "sensor/dht_something.h"`
 * It avoids filename conflicts as you mentioned. For example, I think the `defines.h`, `log.h`, `helpers.h` and possibly `component.h` files could quickly lead to collisions with external libraries too.
   * If external libraries do something like `#include "helpers.h"`, are they guaranteed to always import their own helpers file, or could they accidentally import esphomelib's?
 * It would break all [`esphomeyaml`](https://github.com/OttoWinter/esphomeyaml) installations. When esphomeyaml generates the initial `.cpp` file, it puts `#include "esphomelib/appplication.h"` at the top, which is outside the scope where esphomeyaml writes changes to. Two workarounds I can think of and I could implement if you want:
   * Create a dummy`src/esphomelib/application.h` file that just includes `#include "../application.h"`. Would this work with the Arduino IDE?
   * Alternatively, I could create a small script to the esphomeyaml writer that just replaces all occurrences of `#include "esphomelib/application.h"` with `#include "esphomelib.h"` or `#include "application.h"`.
 * Something I can easily work around, but still: My IDE (CLion) freaks out about relative imports 😥 Anyway, I'm sure I can find a way around this.

And, as I know very little about the Arduino IDE library format:

 * If I want to publish a new release, do I just have to set the `version=` property in `library.properties` to the new value and create a git tag?

per1234:
I've found a much better solution that should address all these excellent points you've raised.

The only problem I'm having is with updating platformio.ini to make the Travis CI build work after moving the example code from the .cpp to .ino files (I don't have much experience with PlatformIO). I'm working on this now and hope to push an update to this PR shortly.

OttoWinter:
Jep, build systems are hard to work with. I usually just spend 2-3 days somehow setting them up through trial and error and never take a look back :)

Thanks a lot for explaining Arduino's build system a bit! It will definitely help me in the future if something is not working/building. The solution of creating a `esphomelib.h` file is also very nice, I must say 👍

Principally, I'm in favor of putting esphomelib onto the Arduino library index, but that would probably also require all used libraries to be on the Arduino library index too - and with FastLED it needs to be very specific version (or rather a specific git tag) in order to work with the ESP32. I don't think that's going to be too easy. So let's leave it with the README instruction, as Arduino IDE users are probably quite used to them anyway 👍

So all in all this PR looks great to me 🎉 Thanks a lot for your effort again! There's currently a merge conflict because I pushed a new commit directly onto the master branch (I should probably stop doing that... but I'm too lazy and there are not too many contributors :P ). I'll resolve it tomorrow, now it's time to sleep for me 💤

per1234:
I have resolved the merge conflict via a force push.

OttoWinter:
Sorry about that. Forgot to fix the conflict...

Anyway, thank you so much for this!

PR #76, opened 2018-06-09
WIP: Add support for SNTP and RTC

### LINKED ISSUE #65: [Feature Request] SNTP Support
Now that I start looking at using Lambda code for more things, I'm realizing it might be nice to have the local RTC set properly.

It appears that both ESP32 and ESP8266 have support for SNTP via the configTime() function:
* https://github.com/espressif/arduino-esp32/blob/master/cores/esp32/esp32-hal-time.c#L46
* https://github.com/esp8266/Arduino/blob/master/cores/esp8266/time.c#L58

As far as exposing this through esphomeyaml, it seems like there are only three things that can be configured:
* GMT Offset (time zone)
* DST offset (if DST is to be observed)
* Server list (1-3 servers)

It would probably be reasonable to have GMT and DST offset default to zero, which effectively puts the RTC at UTC. Similarly, it could default to three pool servers (_n_.pool.ntp.org)

brandond:
Fixes #65 

I'm not much of a C/C++ coder - I usually stick to Python/Java/C#, so I'm looking for feedback on the API and code style. Once that's nailed down I'll put in PRs for yaml and docs.

Note that the RTC component is mostly just a wrapper around common `time.h` stuff. I might pick up an external RTC just so that I have a better feeling for how folks might want to interact with one of these.

### DIFF
--- a/src/esphomelib/application.h
+++ b/src/esphomelib/application.h
@@ -76,6 +76,8 @@
 #include "esphomelib/switch_/switch.h"
 #include "esphomelib/switch_/template_switch.h"
 #include "esphomelib/status_led.h"
+#include "esphomelib/time/rtc_component.h"
+#include "esphomelib/time/sntp_component.h"
 #include "esphomelib/web_server.h"
 
 ESPHOMELIB_NAMESPACE_BEGIN
@@ -692,7 +694,16 @@ class Application {
                                              uint32_t update_interval = 15000);
 #endif
 
+#ifdef USE_RTC_COMPONENT
+  time::RTCComponent *make_rtc_component(const std::string &tz = "UTC");
+#endif
 
+#ifdef USE_SNTP_COMPONENT
+  time::SNTPComponent *make_sntp_component(const std::string &server_1 = "0.pool.ntp.org",
+		                           const std::string &server_2 = "1.pool.ntp.org",
+					   const std::string &server_3 = "2.pool.ntp.org",
+                                           const std::string &tz = "UTC");
+#endif
 
 
 
--- a/src/esphomelib/time/rtc_component.cpp
+++ b/src/esphomelib/time/rtc_component.cpp
@@ -1,0 +1,95 @@
+//
+//  rtc.cpp
+//  esphomelib
+//
+//  Created by Brad Davidson on 06.06.18.
+//  Copyright © 2018 Otto Winter. All rights reserved.
+//
+
+#include "esphomelib/log.h"
+#include "esphomelib/time/rtc_component.h"
+
+#ifdef USE_RTC_COMPONENT
+
+ESPHOMELIB_NAMESPACE_BEGIN
+
+namespace time {
+
+static const char *TAG = "rtc";
+
+EsphomelibTime convert_time_(struct tm *c_tm, time_t c_time){
+  return EsphomelibTime {
+    .second = c_tm->tm_sec,
+    .minute = c_tm->tm_min,
+    .hour = c_tm->tm_hour,
+    .day_of_week = c_tm->tm_wday + 1,
+    .day_of_month = c_tm->tm_mday,
+    .day_of_year = c_tm->tm_yday + 1,
+    .month = c_tm->tm_mon + 1,
+    .year = c_tm->tm_year + 1900,
+    .is_dst = c_tm->tm_isdst,
+    .time = c_time
+  };
+} 
+struct tm revert_time_(const EsphomelibTime &time){
+  struct tm c_tm = tm {
+    .tm_sec = time.second,
+    .tm_min = time.minute,
+    .tm_hour = time.hour,
+    .tm_mday = time.day_of_month,
+    .tm_mon = time.month - 1,
+    .tm_year = time.year - 1900,
+    .tm_wday = time.day_of_week - 1,
+    .tm_yday = time.day_of_year - 1,
+    .tm_isdst = time.is_dst
+  };
+  return c_tm;
+}
+RTCComponent::RTCComponent(const std::string &tz){
+  setenv("TZ", tz.c_str(), 1);
+}
+void RTCComponent::setup(){
+  ESP_LOGCONFIG(TAG, "Setting up RTC...");
+  tzset();
+}
+void RTCComponent::set_timezone(const std::string &tz){
+  setenv("TZ", tz.c_str(), 1);
+  tzset();
+}
+std::string RTCComponent::get_timezone(){
+  const char *tz = getenv("TZ");
+  if (tz == nullptr){ 
+    return {};
+  } else {
+    return std::string(tz);
+  }
+}
+EsphomelibTime RTCComponent::now(){
+  time_t t = ::time(nullptr);
+  struct tm *c_tm = ::localtime(&t);
+  return convert_time_(c_tm, t);
+}
+EsphomelibTime RTCComponent::utcnow(){
+  time_t t = ::time(nullptr);
+  struct tm *c_tm = ::gmtime(&t);
+  return convert_time_(c_tm, t);
+}
+std::string RTCComponent::strftime(const std::string &format, const EsphomelibTime &time){
+  assert(format.size() > 0);
+  std::string timestr;
+  timestr.resize(format.size() * 4);
+  struct tm c_tm = revert_time_(time);
+  size_t len = ::strftime(&timestr[0], timestr.size(), format.c_str(), &c_tm);
+  while (len == 0){
+    timestr.resize(timestr.size() * 2);
+    len = ::strftime(&timestr[0], timestr.size(), format.c_str(), &c_tm);
+  }
+  timestr.resize(len);
+  return timestr;
+}
+
+} // namespace time
+
+ESPHOMELIB_NAMESPACE_END
+
+#endif //USE_RTC_COMPONENT
--- a/src/esphomelib/time/rtc_component.h
+++ b/src/esphomelib/time/rtc_component.h
@@ -1,0 +1,84 @@
+//
+//  rtc_component.h
+//  esphomelib
+//
+//  Created by Brad Davidson on 06.06.18.
+//  Copyright © 2018 Otto Winter. All rights reserved.
+//
+
+#ifndef ESPHOMELIB_TIME_RTC_COMPONENT_H
+#define ESPHOMELIB_TIME_RTC_COMPONENT_H
+
+#include "esphomelib/component.h"
+#include "esphomelib/defines.h"
+#include <stdlib.h>
+#include <time.h>
+
+#ifdef USE_RTC_COMPONENT
+
+ESPHOMELIB_NAMESPACE_BEGIN
+
+namespace time {
+
+/// A more user-friendly version of struct tm from time.h
+/// /note second is generally 0-59; the extra range is to accomodate leap seconds.
+struct EsphomelibTime {
+  /// seconds after the minute [0-61]
+  int second;
+  /// minuntes after the hour [0-59]
+  int minute;
+  /// hours since midnight [0-23]
+  int hour;
+  /// day of the week; sunday=1 [1-7]
+  int day_of_week;
+  /// day of the month [1-31]
+  int day_of_month;
+  /// day of the year [1-366]
+  int day_of_year;
+  /// month; january=1 [1-12]
+  int month;
+  /// year
+  int year;
+  /// daylight savings time flag
+  int is_dst;
+  /// unix epoch time (seconds since UTC Midnight January 1, 1970)
+  time_t time;
+};
+
+
+/// The RTC component exposes common timekeeping functions via the device's local real-time clock.
+///
+/// \note
+/// The C library (newlib) available on ESPs only supports TZ strings that specify an offset and DST info;
+/// you cannot specify zone names or paths to zoneinfo files.
+/// \see https://www.gnu.org/software/libc/manual/html_node/TZ-Variable.html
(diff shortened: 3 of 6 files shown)


brandond:
Right now the C++ interface looks something like this:

```c++
  auto *rtccomponent = App.make_rtc_component("UTC");
  auto *sntpcomponent = App.make_sntp_component();
  sntpcomponent->set_servers({"0.pool.ntp.org", "1.pool.ntp.org"});

  auto time = rtccomponent->time(nullptr);
  auto tm = rtccomponent->localtime(&time);
  ESP_LOGCONFIG("main", "RTC time is %ld (year=%d month=%d day=%d)", time, tm->tm_year, tm->tm_mon, tm->tm_mday);
  ESP_LOGCONFIG("main", "Formatted time is %s", rtc->strftime("%Y-%m-%dT%H:%M:%SZ", tm).c_str());
```

I decided to stick with the classic `time_t` and `struct tm`; hopefully that's not too weird for anyone.

src/esphomelib/application.h
```diff
@@ -663,7 +665,13 @@ class Application {
   MakeESP32HallSensor make_esp32_hall_sensor(const std::string &name, uint32_t update_interval = 15000);
 #endif
 
+#ifdef USE_RTC_COMPONENT
+time::RTCComponent *make_rtc_component(const std::string &tz);
```
OttoWinter: Indent the `make_rtc_component` and `make_sntp_component` component lines please (but not the preprocessor macros)

src/esphomelib/application.h
```diff
@@ -663,7 +665,13 @@ class Application {
...
+#endif
 
+#ifdef USE_SNTP_COMPONENT
+time::SNTPComponent *make_sntp_component();
```
OttoWinter: Why not have the constructor pass in the time servers? Then we could a) set defaults and b) make sure the component is in a valid state when it's used (so that nobody can try to use this component without any servers).

```cpp
time::SNTPComponent *make_sntp_component(std::string server_1 = "1.pool.ntp.org", std::string server_2 = "2.pool.ntp.org", std::string server_3 = "3.pool.ntp.org");
```

src/esphomelib/time/rtc_component.cpp
```diff
@@ -0,0 +1,58 @@
...
+
+RTCComponent::RTCComponent(const std::string &tz){
+  this->set_timezone(tz);
+}
+void RTCComponent::setup(){
+}
+void RTCComponent::set_timezone(const std::string &tz){
+  this->tz_ = tz;
+  setenv("TZ", this->tz_.c_str(), 1);
+  tzset();
```
OttoWinter: This `setenv` and `tzset` code should go into a `setup()` method. esphomelib's API tries to pretty much everywhere it's possible store information in class members. And then only when `setup()` is executed it initializes the hardware.

```cp
RTCComponent::RTCComponent(const std::string &tz)
  : tz_(tz) {

}

void RTCComponent::setup() {
  setenv("TZ", this->tz_.c_str(), 1);
  tzset();
}
```

src/esphomelib/time/rtc_component.h
```diff
@@ -0,0 +1,58 @@
...
+ public:
+  explicit RTCComponent(const std::string &tz);
+
+  void setup() override;
+
+  void set_timezone(const std::string &tz);
+
+  std::string get_timezone();
+
+  time_t time(time_t *_timer);
```
OttoWinter: * These methods should be documented. Also their names are not too great (I know they're from libc, but we don't have to repeat their mistakes). 
* Also, concerning the API, the user shouldn't need to do this:

```cpp
auto time = rtccomponent->time(nullptr);
auto tm = rtccomponent->localtime(&time);
``

I think the `time` method should be dropped and `localtime` should call `time` internally. Also, let's not return a pointer below. lambda users shouldn't need to learn the whole arrow operator syntax. The C standard library's time.h is just a mess and please let's not replicate that here.

```cpp
// Pseudocode
struct EsphomelibTime {
  uint8_t seconds;
  uint8_t minutes;
  uint8_t hours;
  uint8_t day_of_week;
  uint8_t day_of_month;
  uint16_t day_of_year;
  uint8_t month;
  uint16_t year;
}

// ...

// I'm calling them like python's datetime methods here to avoid confusion with libc's API
// Few people will call the utcnow() method anyway, as for most use cases I think now would make
// much more sense (of course for other software this would be different).
EsphomelibTime utcnow() {
  time_t *t = ::time(nullptr);
  struct tm *c_time = ::gmtime(t);
  return prettify_time_(c_time);
}

EsphomelibTime now() {
  time_t *t = ::time(nullptr);
  struct tm *c_time = ::localtime(t);
  return prettify_time_(c_time);
}



EsphomelibTime prettify_time_(struct tm *c_time) {
  return EsphomelibTime {
    .seconds = c_time->tm_sec,
	.minutes = c_time->tm_min,
	.hours = c_time->tm_hour,
	.day_of_week  = c_time->tm_wday,
	.day_of_month = c_time->tm_mday,
	.day_of_year = c_time->tm_yday,
	.month = c_time->tm_mon,
	.year = 1900 + c_time->tm_year,
  };
}
```

Problem then of course is how to handle `strftime`, but I think we can just re-convert then.

src/esphomelib/time/sntp_component.h
```diff
@@ -0,0 +1,52 @@
...
+
+#include "esphomelib/component.h"
+#include "esphomelib/defines.h"
+
+#ifdef ARDUINO_ARCH_ESP32
+  #include "apps/sntp/sntp.h"
+#endif
+#ifdef ARDUINO_ARCH_ESP8266
+  #include "sntp.h"
+#endif
```
OttoWinter: If it's not used in the header file, this include should go into the `.cpp` file.

src/esphomelib/time/sntp_component.h
```diff
@@ -0,0 +1,52 @@
...
+
+  /// Call set_servers with the DNS names of one to three hosts to be used for timekeeping.
+  /// /note
+  /// SNTP is inactive until this method returns; timekeeping should begin within several seconds.
+  void set_servers(const std::vector<std::string> &servers);
+
+  float get_setup_priority() const override;
+};
+
+extern SNTPComponent *global_sntp_component;
```
OttoWinter: Where is this used?

src/esphomelib/time/sntp_component.cpp
```diff
@@ -0,0 +1,59 @@
...
+  sntp_setoperatingmode(SNTP_OPMODE_POLL);
+  #endif
+  #ifdef ARDUINO_ARCH_ESP8266
+  sntp_stop();
+  #endif
+
+  for (uint8_t i = 0; i < servers.size(); i++){
+    sntp_setservername(i, strdup(servers[i].c_str()));
+  }
+  sntp_init();
```
OttoWinter: This initialization code should also go into the `setup()` method. The SNTP servers should be stored by three class members.

```cpp
std::string server_1_;
std::string server_2_;
std::string server_3_;
```

(Also, vectors can have a big overhead, because we know we will only have 3 max, we should declare these as separate variables)

OttoWinter (commented):
Concerning the API, I think `SNTPComponent` should be a subclass of `RTCComponent`. That would make initialization easier and would also make the esphomeyaml implementation a bit easier

brandond:
OK, updated and ready for further review.

src/esphomelib/application.h
```diff
@@ -682,7 +684,16 @@ class Application {
                                              uint32_t update_interval = 15000);
 #endif
 
+#ifdef USE_RTC_COMPONENT
+  time::RTCComponent *make_rtc_component(const std::string &tz = "UTC");
```
OttoWinter: So do I understand correctly that an RTC component *without* sntp/other time providers just reports the uptime?

src/esphomelib/time/rtc_component.h
```diff
@@ -0,0 +1,67 @@
...
+  int minute;
+  int hour;
+  int day_of_week;
+  int day_of_month;
+  int day_of_year;
+  int month;
+  int year;
+  int is_dst;
+  time_t time;
+};
```
OttoWinter: For the methods that `lambda` users would use, I think we should definitely have docs (I know this is a WIP PR, but just a heads up😉). For *example*:

```cpp
struct EsphomelibTime {
  /// The second of the minute from 0 to 59.
  int second;
  /// The minute of the hour from 0 to 59.
  int minute;
  /// The hour of the day from 0 to 23.
  int hour;
  /// The day of the week from 1 (sunday) to 7 (saturday).
  int day_of_week;
  /// The day of the month from 1 to 31.
  int day_of_month;
  /// The day of the year from 1 to 366.
  int day_of_year;
  /// The month of the year from 1 (January) to 12 (December)
  int month;
  /// The year, for example 2018
  int year;
  /// Whether daylights saving's time is active for this time zone at the moment.
  int is_dst;
  /// The underlying C time structure <--- Is this required? It doesn't seem to be used anywhere and essentially just doubles the size of the struct.
  time_t time;
};
```

src/esphomelib/time/sntp_component.cpp
```diff
@@ -0,0 +1,79 @@
+//
+//  ntp_component.cpp
```
OttoWinter: `sntp_component` ❤️

OttoWinter (commented):
Looks pretty much good to me 👍 Just a small comment regarding docs.

For the esphomeyaml side/schema: So you had this configuration format in mind if I understand correctly:

```yaml
rtc:
  platform: (hardware|pcf8523|ds3231|etc)
  timezone: UTC
  id: local_clock
ntp:
  rtc: local_clock
  servers:
    - 0.pool.ntp.org
    - 1.pool.ntp.org
    - 2.pool.ntp.org
```

So then the `rtc` section would configure how the time is *stored* and the *ntp* would act as a way of *providing* the time? I understand why there would be separate `ntp` and `rtc` sections in the configuration, but I think it could be a bit confusing, especially because realistically users would only a single time "store" (pcf8523, ...) and a single time "truth" (only one I can think of is NTP).

I mean if NTP is available, I should be used no matter what and the other stores should only act as "backups". Additionally I think since esphomelib is made for use with a WiFi connection, I think it's safe to assume that everyone using time will also use NTP.

Also I'm not a huge fan of the `rtc` and `ntp` names; technically they're correct but I think bringing it under a common `time` section would be more descriptive.

So maybe at the moment, something like this

```yaml
time:
  timezone: UTC
  id: my_time
  
  # Defaults
  servers:
    - 0.pool.ntp.org
    - 1.pool.ntp.org
    - 2.pool.ntp.org  
```

Which could later be expanded to something like:

```yaml
time:
  timezone: UTC
  id: my_time
  
  external_rtc:
    pcf8523:
      address: 0x42
  
  # Defaults
  servers:
    - 0.pool.ntp.org
    - 1.pool.ntp.org
    - 2.pool.ntp.org
```

I have never worked with arduino/esp projects using any real time clock so I might very well be missing something here, just thought having `rtc` and `ntp` in a config would be a bit confusing.

src/esphomelib/application.h
brandond: My understanding is that (on the ESP32 at least) the RTC lives in the ULP. The memory is cleared along with everything else on boot. It is possible to keep the ULP running while the rest of the system (including main memory) is shut down, in which case the main processor registers and main memory are both lost. But yeah, for most cases, the RTC resets to 0 on boot.

There are a number of battery-backed RTCs available; I was thinking that it might be nice to transparently use and NTP sync those instead of the system RTC if one is available, but that might not be easy to do.

src/esphomelib/time/rtc_component.h
brandond: Will do. I'm not super familiar with doxygen so I appreciate the pointers on where/how to comment such that it will show up in the docs.

brandond: Oh, and that's `time_t` (epoch time) not `struct tm`. It's nice to have in there so you can do basic time math (total seconds between events, etc).

brandond:
If this looks good I'll start on the yaml and docs

OttoWinter:
LGTM 👍

src/esphomelib/time/rtc_component.cpp
```diff
@@ -0,0 +1,95 @@
...
+  assert(format.size() > 0);
+  std::string timestr;
+  timestr.resize(format.size() * 4);
+  struct tm c_tm = revert_time_(time);
+  size_t len = ::strftime(&timestr[0], timestr.size(), format.c_str(), &c_tm);
+  while (len == 0){
+    timestr.resize(timestr.size() * 2);
+    len = ::strftime(&timestr[0], timestr.size(), format.c_str(), &c_tm);
+  }
+  timestr.resize(len);
```
OttoWinter: From my testing these `resize`s quickly lead to a very fragmented heap. And as the ESP8266 does not have any de-fragmentation mechanisms, using `std::string` here leads to frequent failed heap allocations.

I will create a copy of this method in the display renderer using a constant-sized (64 bytes should be plenty) char array allocated on the stack which should eliminate this problem.

OttoWinter (approved):
Ok, so I've been working a lot on implementing displays (LCDs, E-Paper, 8seg, ...) in esphomelib a lot in the last few days and having RTC there would be awesome :)

So I will merge this PR, modify it a bit for the display rendering code, and write the esphomeyaml hooks so that people can start using it with the next release.

Thank you for all this work on RTC&SNTP!

PR #111, opened 2018-07-24
Fix panic in binary_sensor template

nunofgs:
Fixes a panic with the binary_sensor due to a typo in the lambda declaration.

* Fixes [esphomeyaml#74](https://github.com/OttoWinter/esphomeyaml/issues/74).


nunofgs:
@OttoWinter FYI, tests are failing but it seems unrelated. I'll rebase this branch once _master_ is fixed.

OttoWinter:
Ah yes, thanks. I was away for a while so thanks very much for the fix. Don't worry about the tests :)

PR #115, opened 2018-07-29
output: configurable default pin state

jkoelker:
Adds the `default_pin_high` paramater to `GPIOBinaryOutputComponent`'s
constructor with the default value of false. Allows keeping the pin high
if needed during `setup()` (e.g. when using an enable low relay
board).


OttoWinter:
The way to do this is to pass the constructor an inverted GPIO pin like this:

```cpp
App.make_gpio_output(GPIOOutputPin(2, OUTPUT, true));
```

All outputs are off by default because handling inversion/default pin mode should be handled on another layer (in this case in the `GPIOOutputPin`). If each layer (for example: `GPIOOutputPin`, `output`, `light`) would have their own inversion settings, it would very quickly become very complicated.

jkoelker:
Excellent, thanks, that's much better.

PR #116, opened 2018-07-29
[RFC] application: add critical device mode

jkoelker:
Adds the boolean flag `critical_device` to `Application::setup()` which
defaults to false. If set to true, setup will not block on the
components `can_proceed()`, allowing the device to transition to
`loop()` as soon as possible.

My particular use case is to use `esphomelib` to drive a refrigerator keeping it around 10c for cheese aging.  To do this I'd like to be able to run my logic as fast as possible on startup (potentially after a power loss), the blocking on `can_proceed()` would prevent this if the mqtt host or the wifi was down. It appears that the connections still get established when needed during the `loop()` or fails fast, allowing for other logic to run to trigger the relays as needed.

### DIFF
--- a/src/esphomelib/application.cpp
+++ b/src/esphomelib/application.cpp
@@ -44,7 +44,7 @@ using namespace esphomelib::remote;
 
 static const char *TAG = "application";
 
-void Application::setup() {
+void Application::setup(bool critical_device) {
   ESP_LOGI(TAG, "Running through setup()...");
   assert(this->application_state_ == COMPONENT_STATE_CONSTRUCTION && "setup() called twice.");
   ESP_LOGV(TAG, "Sorting components by setup priority...");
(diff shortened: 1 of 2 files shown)


OttoWinter:
Uhm that's pretty much exactly what [`get_setup_priority`](https://esphomelib.com/api/core/component.html#_CPPv3NK9Component18get_setup_priorityEv) is for. If you're implementing a custom component, you just have to override that method and return a high value like so:

```cpp
// Pseudo-Code
class MyCustomComponent : public Component {
  float get_setup_priority() override {
    return 100.0f;
  }
```

This way, the `setup()` and `loop()` functions will be executed before anything non-vital. And if you want to change the setup priority of a built-in component (for example DHTComponent), you just have to override it in a custom component:

```cpp
class MyDHTComponent : public DHTComponent {
  float get_setup_priority() override { return 100.0f; }
}
```

----

The whole setup priority exists so that there cannot be any situations where dependencies are not set up when used. For example a sensor sending data via MQTT while MQTT is not set up. As such, the proposed `critical_device` flag in this PR probably breaks a bunch of stuff in certain situations and the setup priority model should be used instead.

jkoelker:
Good deal, I was hoping to avoid a custom component and just stuffing a bit of logic in the `loop()`, I now see in the `setup()` do-while loop, that the `loop()` of all the previous components is called.

PR #123, opened 2018-08-09
Run travis jobs in parallel

lobradov:
This will enable Travis to run different jobs in parallel. 

It speeds up overall testing (from 9 to 3 min), and provides per-job view of status, for easier problem finding.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,3 +23,5 @@ cmake-build-release/
 
 # pyenv
 .python-version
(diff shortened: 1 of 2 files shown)


lobradov:
Actually, all examples compile in 15min in sequential and in 3min in parallel order. 
Adding more examples doesn't increase time of tests.

OttoWinter:
Thank you!

PR #124, opened 2018-08-09
Fix examples

lobradov:
Two examples are broken today and they don't compile.
This causes Travis to show that lib build is broken.


janpieper:
I would suggest to change

> [...] otherwise, compiled will complain [...]

to

> [...] otherwise, compilation will complain [...]

lobradov:
True, autocorrect changed "compiler" to "compiled".
I'll change.

OttoWinter:
Thanks 😺

PR #141, opened 2018-09-07
Add Samsung IR codes

escoand:
Added Samsung IR codes. Receiving successfully tested with UE40ES5700.

### DIFF
--- a/src/esphomelib/remote/samsung.cpp
+++ b/src/esphomelib/remote/samsung.cpp
@@ -1,0 +1,95 @@
+//
+//  samsung.cpp
+//  esphomelib
+//
+//  Created by Otto Winter on 06.06.18.
+//  Copyright © 2018 Otto Winter. All rights reserved.
+//
+
+#include "esphomelib/defines.h"
+
+#ifdef USE_REMOTE
+
+#include "esphomelib/remote/samsung.h"
+#include "esphomelib/log.h"
+
+ESPHOMELIB_NAMESPACE_BEGIN
+
+namespace remote {
+
+#ifdef USE_REMOTE_RECEIVER
+static const char *TAG = "remote.samsung";
+#endif
+
+static const uint32_t HEADER_HIGH_US = 4500;
+static const uint32_t HEADER_LOW_US = 4500;
+static const uint32_t BIT_HIGH_US = 560;
+static const uint32_t BIT_ONE_LOW_US = 560;
+static const uint32_t BIT_ZERO_LOW_US = 1690;
+
+#ifdef USE_REMOTE_TRANSMITTER
+SamsungTransmitter::SamsungTransmitter(const std::string &name, uint32_t data, uint8_t nbits)
+    : RemoteTransmitter(name), data_(data), nbits_(nbits) {}
+
+void SamsungTransmitter::to_data(RemoteTransmitData *data) {
+  data->set_carrier_frequency(38000);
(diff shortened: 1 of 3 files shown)


escoand:
This will be one step for #141

src/esphomelib/remote/samsung.cpp
```diff
@@ -0,0 +1,102 @@
...
+static const uint32_t FOOTER_HIGH_US = 560;
+static const uint32_t FOOTER_LOW_US = 560;
+
+#ifdef USE_REMOTE_TRANSMITTER
+SamsungTransmitter::SamsungTransmitter(const std::string &name, uint32_t data, uint8_t nbits)
+    : RemoteTransmitter(name), data_(data), nbits_(nbits) {}
+
+void SamsungTransmitter::to_data(RemoteTransmitData *data) {
+  data->set_carrier_frequency(38000);
+  data->reserve(2 + this->nbits_ * 2u);
```
OttoWinter: This should be `4 + this->nbits_ * 2u`, as there is also a footer.

src/esphomelib/remote/samsung.cpp
```diff
@@ -0,0 +1,102 @@
...
+    if (data->expect_item(BIT_HIGH_US, BIT_ONE_LOW_US)) {
+      *data_ = (*data_ << 1) | 1;
+    } else if (data->expect_item(BIT_HIGH_US, BIT_ZERO_LOW_US)) {
+      *data_ = (*data_ << 1) | 0;
+    } else {
+      return false;
+    }
+  }
+
+  //if (!data->expect_item(FOOTER_HIGH_US, FOOTER_LOW_US))
```
OttoWinter: Please remove this commented line. Comments should be for sentences documenting/explaining the code, and old outdated fragments should not remain in the codebase. (Probably I've missed this rule somewhere, but it's good practice)

src/esphomelib/remote/samsung.cpp
```diff
@@ -0,0 +1,102 @@
...
+  if (!data->expect_mark(FOOTER_HIGH_US))
+    return false;
+
+  return true;
```
OttoWinter: While we're at it:

```cpp
return data->expect_mark(FOOTER_HIGH_US);
```

src/esphomelib/remote/samsung.cpp
```diff
@@ -0,0 +1,102 @@
...
+}
+#endif
+
+#ifdef USE_REMOTE_RECEIVER
+bool decode_samsung(RemoteReceiveData *data, uint32_t *data_, uint8_t *nbits) {
+  if (!data->expect_item(HEADER_HIGH_US, HEADER_LOW_US))
+    return false;
+
+  *data_ = 0;
+  for (*nbits = 0; *nbits < 32; (*nbits)++) {
```
OttoWinter: The decoder doesn't need the `nbits` parameter as it will always be 32 after a successful decode. In the LG class, this was required for checking if a 28-bit or 32-bit code was used, but here it doesn't really make sense (at least for *decoding*).

OttoWinter (commented):
Thanks! Just a few style comments.

escoand:
Fixed your comments.

Maybe there is also some kind of address and command. Every button on my remote starts with 0xE0E0. But this shouldn't matter here...

OttoWinter (approved):
LGTM. Thanks! 😺

Issue #128, opened 2018-08-16
DHT11 only working without checksum verification


janpieper:
I have a ESP-12E, a DHT11 (3 pins) sensor and the following configuration:

```
- sensor:
  - platform: dht
    model: DHT11
    pin: GPIO2
    temperature:
      name: "Temperature"
    humidity:
      name: "Humidity"
```

The wiring:

* (ESP-12E) GND - GND (DHT11)
* (ESP-12E) 3V3 - VCC (DHT11)
* (ESP-12E) D4 - DATA (DHT11)

When checking the logs, I see errors and warnings popping up because the checksum verification fails :disappointed: 

```
...
[08:18:38][E][sensor.dht:151]: Checksum invalid: 71!=80
[08:18:38][W][sensor.dht:059]: Invalid readings! Please check your wiring (pull-up resistor, pin_ number) and consider manually specifying the DHT model using the model option. Error code: 0
[08:18:53][VV][sensor.dht:141]: Data: Hum=0b0010111100000000, Temp=0b0001100000001001, Checksum=0b01010000
[08:18:53][E][sensor.dht:151]: Checksum invalid: 71!=80
[08:18:53][W][sensor.dht:059]: Invalid readings! Please check your wiring (pull-up resistor, pin_ number) and consider manually specifying the DHT model using the model option. Error code: 0
[08:19:08][VV][sensor.dht:141]: Data: Hum=0b0100000100000000, Temp=0b0001100100000100, Checksum=0b01011110
[08:19:08][E][sensor.dht:151]: Checksum invalid: 90!=94
[08:19:08][W][sensor.dht:059]: Invalid readings! Please check your wiring (pull-up resistor, pin_ number) and consider manually specifying the DHT model using the model option. Error code: 0
[08:19:23][VV][sensor.dht:141]: Data: Hum=0b0100110000000000, Temp=0b0001101000000111, Checksum=0b01101101
[08:19:23][E][sensor.dht:151]: Checksum invalid: 102!=109
[08:19:23][W][sensor.dht:059]: Invalid readings! Please check your wiring (pull-up resistor, pin_ number) and consider manually specifying the DHT model using the model option. Error code: 0
[08:19:38][VV][sensor.dht:141]: Data: Hum=0b0011011000000000, Temp=0b0001101100000011, Checksum=0b01010100
[08:19:38][E][sensor.dht:151]: Checksum invalid: 81!=84
[...]

PR #142, opened 2018-09-10
Fix checksum verification for DHT11 sensors

janpieper:
This PR is meant as a fix for #128. See the issue for more details.

I've found [TinyDHT.cpp @ adafruit/TinyDHT](https://github.com/adafruit/TinyDHT/blob/103fc3f423e71087aaae1a2e8267b9d54fcc6da0/TinyDHT.cpp#L155-L159), which uses a different implementation to check if the read data is valid. It works fine with my DHT11 sensors.

**Attention:** This change is **NOT** tested with DHT21 and DHT22 sensors!


OttoWinter:
See https://github.com/OttoWinter/esphomelib/issues/128#issuecomment-423093169, apparently this issue is a bit more complicated. But I have written a small fix locally (+ improved some of the error messages).

OttoWinter:
Fix is in 1.8.0

PR #151, opened 2018-09-25
Project Logos

messismore:
esphomelib-logo-specimen

I touched up logo no. 3 from [this thread](https://github.com/OttoWinter/esphomeyaml/issues/53).

The assets include the logo with both the -lib and the -yaml name as well as  just the bare logo each as SVG and PNG. The typeface is [Arvo](https://fonts.google.com/specimen/Arvo) which is available under the [Open Font License](https://scripts.sil.org/cms/scripts/page.php?site_id=nrsi&id=OFL_web).

I am not really sure where these should go, hence the ‚assets’ folder. I’ll gladly change the PR to what you think works best.


OttoWinter:
Thank you so much!!! And sorry, I should have mentioned: The logo should go into the `_static` folder in the docs repo: https://github.com/OttoWinter/esphomedocs/tree/current/_static. I've already set up the necessary configuration options for sphinx (the docs engine) so you only need to replace these files (I generated these with https://realfavicongenerator.net/, but feel free to use something else):

* logo-full.png - The logo with "esphomelib" text
* logo.png - The logo without any text
* favicon.ico - logo.png, just in another (ancient) format

Feel free to drop in other images there too (SVG versions, and the esphomeyaml logo) so that they can be used in the future.

PR #152, opened 2018-09-25
Add toggle payload to MQTT switch component

### LINKED ISSUE #138: MQTT command to toggle a component
I would like to toggle a light of an other esphomelib device. This is currently just possible when subscribing to the state topic of the destination device but not with a toggle MQTT telegram.
What do you think about this additional command?

OttoWinter:
When payload "TOGGLE" is sent to command_topic, it will now toggle the state of the switch.

Fixes #138


escoand:
This is not working for me. The command was recognized also the "starting transition" message was logged but the state is not switched

OttoWinter:
@escoand This PR only applies to switches, not lights (I'm assuming you're talking about a light because of `starting transition`)

escoand:
Are you willing to do this for lights as well?

PR #169, opened 2018-10-01
Fix light color not changing when also changing effect

### LINKED ISSUE #166: Turning effect off freezes light in effect state until change
Fix for this issue: #169

(Edit 2: now also tried on digital strips, and it occurs on all effects that are present)

When turning off an effect, the lights freeze in the current state (e.g. Strobe remains white // off, when turning the effect to none) until you change one of the light values (brightness, color,...)
(Hass still reports the proper values, which the light should have)

(Board: NodeMCUv2, esp8266, hassio esphomeyaml addon "bleeding-edge" -> currently 1.8.1)
(Also, the strobe effect always uses the white color but I guess that is intended?)

Relevant config entry:
```
output:
  - platform: esp8266_pwm
    pin: D0
    id: red_output
  - platform: esp8266_pwm
    pin: D4
    id: green_output
  - platform: esp8266_pwm
    pin: D8
    id: blue_output

light:
  - platform: rgb
    name: "led_1_strip"
    id: leds
    red: red_output
    green: green_output
    blue: blue_output
    gamma_correct: 1.0
    effects:
      - random:
      - strobe:
      - flicker:
```

(Edit 1: When the strobe effect currently displays the bright phase, and ```{"state": "ON", "effect": "None"}``` is then called, it stops in the white phase.)

### LINKED ISSUE #167: Not resetting effect to None, when using home assistant scenes
Fix for this issue: #169 
(Board: NodeMCUv2, esp8266, hassio esphomeyaml addon "bleeding-edge" -> currently 1.8.1)

Home Assistant scene which are not working as expected:
```
- name: Example_Scene_Name_NOT_Working_1
  entities:
    light.led_1_strip:
      state: on
      rgb_color: [255,0,0]
      effect: None
      brightness: 200
```
only changes the effect to "None" but does not change the colors.

```
- name: Example_Scene_Name__NOT_Working_2
  entities:
    light.led_1_strip:
      state: on
      rgb_color: [255,0,0]
      effect: Flicker"
      brightness: 200
```
also only changes the effect to "Flicker" but does not change the colors.
(Even if we're changing back to None through e.g. the hass interface)

```{"state": "ON", "effect": "None"}``` works and turns the effect off, as expected.
```{"state": "ON", "color": {"r": 0, "g": 0, "b": 255}, "brightness": 200, "effect": "None"}``` only changes the effect to "None".
```{"state": "ON", "color": {"r": 0, "g": 0, "b": 255}, "brightness": 200, "effect": "Flicker"}``` only changes the effect to "Flicker".

Relevant esphomeyaml config entry:
```
light:
  - platform: rgb
    name: "led_1_strip"
    id: leds
    red: red_output
    green: green_output
    blue: blue_output
    gamma_correct: 1.0
    effects:
      - random:
      - strobe:
      - flicker:
```

Log entry (when activating the scene at the top, which doesn't do anything at all):
```
[16:17:16][VV][component:136]: Running defer '':0 with interval=0 last_execution=0 (now=1048895)
[16:17:16][V][light.state:152]: Interpreting light JSON.
[16:17:16][V][light.light_color_values:144]: Parsing light color values JSON.
[16:17:16][V][light.light_color_values:149]:     state=ON
[16:17:16][V][light.light_color_values:155]:     brightness=0.78
[16:17:16][V][light.light_color_values:164]:     r=0.00, g=0.00, b=1.00
```

TheJulianJES:
Designed to fix: #167 but also seems to fix #166 (because we're now starting a transition back to the normal color lights, if we reset the effect.)

Previously:
```{"state": "ON", "color": {"r": 0, "g": 0, "b": 255}, "brightness": 200, "effect": "None"}```
If we had this payload, only the effect would be set (e.g. to none) but not the actual colors provided.
Regardless of how many times this was sent (or if the effect was already "None").
Now we're starting the effect (in case of "None", we're stopping the effect) and are then also setting the colors (e.g. starting the transition).

This is needed because if we stop the effect (setting to "None"), the previous colors from the effect stayed (e.g. if we had been using "Strobe", the colors could either be 255, 255, 255 or 0, 0, 0).
This will fix #166 

In order to fix #167 we also want to stop the effect but also start the transition to the normal colors, not just leave them out, because that would make the whole "color" part of the payload useless, if an effect is provided in the same payload.

So the fix for the issue is the same.
Since I'm not an experienced C++ developer at all (just started for esphomelib actually), I can't really say if this fix is considered "clean" and you might come up with a better solution.
But I can say that this fix was tested and doesn't seem to have any side effects at all.

### DIFF
--- a/src/esphomelib/light/light_state.cpp
+++ b/src/esphomelib/light/light_state.cpp
@@ -160,11 +160,13 @@ void LightState::parse_json(const JsonObject &root) {
   } else if (root.containsKey("transition")) {
     auto length = uint32_t(float(root["transition"]) * 1000);
     this->start_transition(v, length);
-  } else if (root.containsKey("effect")) {
-    const char *effect = root["effect"];
-    this->start_effect(effect);
   } else {
     this->start_default_transition(v);
+  }
+
+  if (root.containsKey("effect")) {
+    const char *effect = root["effect"];
+    this->start_effect(effect);
   }
 }
 


OttoWinter (commented):
Ah yes, good catch!

For the code: I would rather like the `if (root.containsKey("effect")) {` check to be **after** the whole flash/transition/default transition block, not before it. The reason is that in the future, effects could use the color at the start of the effect as a parameter.

TheJulianJES:
Uh, oh. Messed up git now. Will fix in a couple of mins, so this shouldn't be merged atm.

TheJulianJES:
Can also remove the blank line if you want. I left it in (for readability and) so it isn't confused with being part of the "if/else" block.

OttoWinter (approved):
Nah it's fine. Thanks!

PR #186, opened 2018-10-08
Fix LCD display include

OttoWinter:
Fixes OttoWinter/esphomeyaml#154


dennis1987:
Now it works!
Nice Job!
Thank you very much!

PR #287, opened 2018-11-23
Fix template cover spamming logs

OttoWinter:
## Checklist:
  - [x] Check this box if you have read, understand, comply, and agree with the [Code of Conduct](https://github.com/OttoWinter/esphomelib/blob/master/CODE_OF_CONDUCT.md).


postlund:
This should fix 
https://github.com/OttoWinter/esphomeyaml/issues/239, right?

PR #323, opened 2018-12-14
Fix large JSON payloads being cut off

quazzie:
I don't know this codebase well enough so there might be a better way and or other place but this fixes my problem.
When publishing payloads with size about 600-700 bytes it got cut-off.
With this fix i get to send all my payload :)


OttoWinter (approved):
Ah yes, true - the check is failing because of the null terminator.

PR #334, opened 2018-12-26
added tx20 wind sensor

TheEggi:
## Description:
The tx20 is a simple wind sensor, which can be used to get information about wind speed and direction. 
Sensor can be found here: https://www.amazon.de/Technoline-TX-20/dp/B01HXZ3KLA

The sensor must be connected as follows:

Yellow ground (GND)
Green mass (DTR) -> should also go to GND
Red 3.3V (Vcc)
Black GPIO_2 (TxD) -> this one will be configured.

Description of the protocol:
https://www.john.geek.nz/2011/07/la-crosse-tx20-anemometer-communication-protocol/
http://fabrizio.zellini.org/decoding-la-crosse-tx20-anemometer-with-arduino

Implementations:
https://github.com/arendst/Sonoff-Tasmota/blob/development/sonoff/xsns_35_tx20.ino
https://github.com/bunnyhu/ESP8266_TX20_wind_sensor
http://blog.bubux.de/windsensor-tx20-mit-esp8266/

Detailed description of the sensor can be found here: https://github.com/arendst/Sonoff-Tasmota/pull/3146.

**Related issue (if applicable):** 

**Pull request in [esphomeyaml](https://github.com/OttoWinter/esphomeyaml) with YAML changes (if applicable):** OttoWinter/esphomeyaml# https://github.com/OttoWinter/esphomeyaml/pull/275
**Pull request in [esphomedocs](https://github.com/OttoWinter/esphomedocs) with documentation (if applicable):** OttoWinter/esphomedocs#https://github.com/esphome/ESPHome-Docs/pull/150
## Checklist:

  - [x] Documentation added/updated in [esphomedocs](https://github.com/OttoWinter/esphomedocs).


platformio.ini
```diff
@@ -1,7 +1,7 @@
 ; This file is here for developing esphomelib, as a user you don't need this.
 
 [platformio]
-env_default = livingroom
+env_default = tx20sensor
```
OttoWinter: Remove this change.

platformio.ini
```diff
@@ -104,3 +104,19 @@ framework = arduino
...
 build_flags = ${common.build_flags}
 src_filter = ${common.src_filter} +<examples/sonoff-b1/sonoff-b1.cpp>
+
+[env:tx20sensor]
+platform = espressif8266
+board = nodemcuv2
+framework = arduino
+lib_deps = ${common.lib_deps}
+build_flags = ${common.build_flags}
+src_filter = ${common.src_filter} +<examples/tx20-sample/tx20-sample.cpp>
```
OttoWinter: Please only add one of the esp32/esp8266 versions. Each environment makes the test scripts compile all of esphomelib's code base another time.

platformio.ini
```diff
@@ -104,3 +104,19 @@ framework = arduino
...
+
+[env:tx20sensor-esp32]
+platform = espressif32
+board = nodemcu-32s
+framework = arduino
+lib_deps = ${common.lib_deps}
+build_flags = ${common.build_flags}
+src_filter = ${common.src_filter} +<examples/tx20-sample/tx20-sample.cpp>
```
OttoWinter: Missing newline at end of file

src/esphomelib/application.h
```diff
@@ -158,8 +160,7 @@ class Application {
    * @param tx_buffer_size The size of the printf buffer.
    * @return The created and initialized LogComponent.
    */
-  LogComponent *init_log(uint32_t baud_rate = 115200,
-                         size_t tx_buffer_size = 512);
+  LogComponent *init_log(uint32_t baud_rate = 115200, size_t tx_buffer_size = 512);
```
OttoWinter: Please remove these cleanup changes. It makes it hard to review this PR (and makes merge conflicts that much harder to fix). If you want to fix formatting in this file, please do it in another PR.

src/esphomelib/sensor/tx20_component.cpp
```diff
@@ -0,0 +1,160 @@
+// Based on:
```
OttoWinter: Remove this.

src/esphomelib/sensor/tx20_component.cpp
```diff
@@ -0,0 +1,160 @@
...
+
+#include "esphomelib/defines.h"
+
+#ifdef USE_TX20
+
+#include "esphomelib/espmath.h"
+#include "esphomelib/log.h"
+#include "esphomelib/text_sensor/text_sensor.h"
+#include "tx20_component.h"
```
OttoWinter:
```suggestion
#include "esphomelib/sensor/tx20_component.h"
```

src/esphomelib/sensor/tx20_component.cpp
```diff
@@ -0,0 +1,160 @@
...
+
+const char UNIT_KMH[] = "km/h";
+const char ICON_WIND_SPEED[] = "mdi:weather-windy";
+const char ICON_WIND_DIRECTION[] = "mdi:sign-direction";
+const char EMPTY[] = "";
+const uint16_t TX20_BIT_TIME = 1220;   // microseconds
+const uint8_t TX20_RESET_VALUES = 60;  // seconds
+
+const std::vector<std::string> DIRECTIONS = {"N", "NNE", "NE", "ENE", "E", "ESE", "SE", "SSE",
+                                             "S", "SSW", "SW", "WSW", "W", "WNW", "NW", "NNW"};
```
OttoWinter: Don't use `std::vector` here. I believe the compiler will not optimize it away if this integration is not used. Please change to a standard C-style array.

src/esphomelib/sensor/tx20_component.cpp
```diff
@@ -0,0 +1,160 @@
...
+ESPHOMELIB_NAMESPACE_BEGIN
+
+namespace sensor {
+
+static const char *TAG = "sensor.tx20";
+
+const char UNIT_KMH[] = "km/h";
+const char ICON_WIND_SPEED[] = "mdi:weather-windy";
+const char ICON_WIND_DIRECTION[] = "mdi:sign-direction";
+const char EMPTY[] = "";
```
OttoWinter: Move these ICON and UNIT constants to `sensor.h` and `sensor.cpp`

src/esphomelib/sensor/tx20_component.cpp
```diff
@@ -0,0 +1,160 @@
...
+const uint16_t TX20_BIT_TIME = 1220;   // microseconds
+const uint8_t TX20_RESET_VALUES = 60;  // seconds
+
+const std::vector<std::string> DIRECTIONS = {"N", "NNE", "NE", "ENE", "E", "ESE", "SE", "SSE",
+                                             "S", "SSW", "SW", "WSW", "W", "WNW", "NW", "NNW"};
+
+static bool tx20_available = false;
+static float tx20_wind_speed_kmh = 0;
+static uint8_t tx20_wind_direction = 0;
+static bool firstInstance = false;
```
OttoWinter: `first_instance`

src/esphomelib/sensor/tx20_component.cpp
```diff
@@ -0,0 +1,160 @@
...
+                             const std::string &wind_direction_text_name, GPIOPin *pin, uint32_t update_interval)
+    : PollingComponent(update_interval),
+      wind_speed_sensor_(new TX20WindSpeedSensor(wind_speed_name, this)),
+      wind_direction_sensor_(new TX20WindDirectionSensor(wind_direction_name, this)),
+      wind_direction_text_sensor_(new TX20WindDirectionTextSensor(wind_direction_text_name)),
+      pin_(pin) {
+  ESP_LOGV(TAG, "Start TX20Component...");
+  if(firstInstance){
+    ESP_LOGE(TAG, "Multiple TX20 Sensors are not supported!!");
+  }
```
OttoWinter: No logging in constructor please.

src/esphomelib/sensor/tx20_component.cpp
```diff
@@ -0,0 +1,160 @@
...
+void TX20Component::setup() {
+  ESP_LOGCONFIG(TAG, "Setting up TX20...");
+  this->pin_->setup();
+
+  attachInterrupt(this->pin_->get_pin(), TX20Component::start_read_, RISING);
+  global_tx20_ = this;
+}
+void TX20Component::dump_config() {
+  ESP_LOGCONFIG(TAG, "TX20:");
+  LOG_PIN(" Pin:", TX20Component::pin_);
```
OttoWinter:
```suggestion
  LOG_PIN("  Pin:", this->pin_);
```

src/esphomelib/sensor/tx20_component.cpp
```diff
@@ -0,0 +1,160 @@
...
+  uint8_t tx20_sb = 0;
+  uint8_t tx20_sd = 0;
+  uint8_t tx20_se = 0;
+  uint8_t tx20_sc = 0;
+  uint8_t tx20_sf = 0;
+
+  delayMicroseconds(TX20_BIT_TIME / 2);
+
+  for (int bitcount = 41; bitcount > 0; bitcount--) {
+    uint8_t dpin = (digitalRead(this->pin_->get_pin()));
```
OttoWinter:
```suggestion
    uint8_t dpin = this->pin_->digital_read();
```

src/esphomelib/sensor/tx20_component.cpp
```diff
@@ -0,0 +1,160 @@
...
+  chk &= 0xf;
+
+  if ((chk == tx20_sd) && (tx20_sc < 400)) {  // if checksum seems to be ok and wind speed below 40 m/s
+    tx20_available = true;
+  }
+
+  // Must clear this bit in the interrupt register,
+  // it gets set even when interrupts are disabled
+  tx20_wind_direction = tx20_sb;
+  tx20_wind_speed_kmh = float(tx20_sc) * 0.36;
```
OttoWinter:
```suggestion
  tx20_wind_speed_kmh = float(tx20_sc) * 0.36f;
```

src/esphomelib/sensor/tx20_component.h
```diff
@@ -0,0 +1,59 @@
+
```
OttoWinter: Remove blank line at start (will become important in future release)

src/esphomelib/sensor/tx20_component.cpp
```diff
@@ -0,0 +1,160 @@
...
+#ifdef ARDUINO_ARCH_ESP32
+  GPIO_REG_WRITE(GPIO.status_w1tc, 1 << this->pin_->get_pin());
+#endif
+#ifdef ARDUINO_ARCH_ESP8266
+  GPIO_REG_WRITE(GPIO_STATUS_W1TC_ADDRESS, 1 << this->pin_->get_pin());
+#endif
```
OttoWinter: > // Must clear this bit in the interrupt register,
> // it gets set even when interrupts are disabled

Why do we need to clear this flag?

src/esphomelib/sensor/tx20_component.cpp
```diff
@@ -0,0 +1,160 @@
...
+      tx20_sd = tx20_sd >> 1 | ((dpin ^ 1) << 3);
+    } else if (bitcount > 41 - 5 - 4 - 12 - 4 - 4) {
+      // wind dir
+      tx20_se = tx20_se >> 1 | (dpin << 3);
+    } else {
+      // windspeed
+      tx20_sf = tx20_sf >> 1 | (dpin << 11);
+    }
+
+    delayMicroseconds(TX20_BIT_TIME);
```
OttoWinter: So this interrupt will take 41*1220 microseconds (50 milliseconds) to read the wind sensor? That's not good for a real-time system like an ESP - the WiFi controller will not be able to respond to packets during this time and esphomelib's core loop will not continue. Please try re-writing this into an asynchronous version, see the UART sensors for inspiration.

src/esphomelib/sensor/tx20_component.h
```diff
@@ -0,0 +1,59 @@
...
+extern const char EMPTY[];
+
+using TX20WindSpeedSensor =
+    sensor::EmptyPollingParentSensor<1, ICON_WIND_SPEED, UNIT_KMH>;
+using TX20WindDirectionSensor =
+    sensor::EmptyPollingParentSensor<1, ICON_WIND_DIRECTION, EMPTY>;
+using TX20WindDirectionTextSensor = text_sensor::TextSensor;
+
+/// This class implements support for the TX20 WindSpeed+WindDirection sensor.
+class TX20Component : public PollingComponent {
```
OttoWinter: Why make this a `PollingComponent` when the sensor is asynchronous (using interrupts) ?

src/esphomelib/sensor/tx20_component.cpp
```diff
@@ -0,0 +1,160 @@
...
+  }
+
+  uint8_t chk = (tx20_sb + (tx20_sc & 0xf) + ((tx20_sc >> 4) & 0xf) + ((tx20_sc >> 8) & 0xf));
+  chk &= 0xf;
+
+  if ((chk == tx20_sd) && (tx20_sc < 400)) {  // if checksum seems to be ok and wind speed below 40 m/s
```
OttoWinter: Is there a reason wind speeds above 40m/s are discarded?

src/esphomelib/sensor/tx20_component.cpp
```diff
@@ -0,0 +1,160 @@
...
+
+TX20Component::TX20Component(const std::string &wind_speed_name, const std::string &wind_direction_name,
+                             const std::string &wind_direction_text_name, GPIOPin *pin, uint32_t update_interval)
+    : PollingComponent(update_interval),
+      wind_speed_sensor_(new TX20WindSpeedSensor(wind_speed_name, this)),
+      wind_direction_sensor_(new TX20WindDirectionSensor(wind_direction_name, this)),
+      wind_direction_text_sensor_(new TX20WindDirectionTextSensor(wind_direction_text_name)),
+      pin_(pin) {
+  ESP_LOGV(TAG, "Start TX20Component...");
+  if(firstInstance){
```
OttoWinter: You can also set `global_tx20_` here and check if `global_tx20_ != nullptr`

src/esphomelib/sensor/tx20_component.cpp
```diff
@@ -0,0 +1,160 @@
...
+const char EMPTY[] = "";
+const uint16_t TX20_BIT_TIME = 1220;   // microseconds
+const uint8_t TX20_RESET_VALUES = 60;  // seconds
+
+const std::vector<std::string> DIRECTIONS = {"N", "NNE", "NE", "ENE", "E", "ESE", "SE", "SSE",
+                                             "S", "SSW", "SW", "WSW", "W", "WNW", "NW", "NNW"};
+
+static bool tx20_available = false;
+static float tx20_wind_speed_kmh = 0;
+static uint8_t tx20_wind_direction = 0;
```
OttoWinter: Don't make these global variables. Even though only one of these can exist at a time, please make them class members.

src/esphomelib/sensor/tx20_component.cpp
```diff
@@ -0,0 +1,160 @@
...
+  LOG_UPDATE_INTERVAL(this);
+}
+float TX20Component::get_setup_priority() const {
+  return setup_priority::HARDWARE_LATE;
+}
+
+void TX20Component::update() {
+  if (tx20_available) {
+    ESP_LOGV(TAG, "Updating TX20...");
+    std::string direction = DIRECTIONS.at(tx20_wind_direction);
```
OttoWinter: Check if `DIRECTIONS` contains the key first. We do not want to crash the system if we get a faulty reading.

src/esphomelib/sensor/tx20_component.cpp
```diff
@@ -0,0 +1,160 @@
...
+
+    this->wind_direction_sensor_->publish_state(tx20_wind_direction);
+    this->wind_speed_sensor_->publish_state(tx20_wind_speed_kmh);
+    this->wind_direction_text_sensor_->publish_state(direction);
```
OttoWinter: Shouldn't `tx20_available` be reset after this block?

src/esphomelib/sensor/tx20_component.cpp
TheEggi: 51.1 is the maximum that is supported by the sensor. I have not seen mine go anywhere close to 40 - so my guess is that it is not working so well in that range.

platformio.ini
TheEggi: Fixed.

src/esphomelib/sensor/tx20_component.cpp
TheEggi: Correct.

src/esphomelib/sensor/tx20_component.cpp
TheEggi: Any one in particular? Did not quite work out the way as I was intending it to do:
- Intention was to move it to the loop and check the with micros() if the elapsed time is >= as my timeout. Result was not quite working.

src/esphomelib/sensor/tx20_component.cpp
OttoWinter: Ok so the sensor has a checksum, so that should be enough for checking if everything's alright. Maybe somebody else _has_ conditions where the wind sensor gets above 40 m/s (up to the 51.1 max).

Can we just drop the second check? I think the checksum by itself is just fine.

src/esphomelib/sensor/tx20_component.cpp
```diff
@@ -0,0 +1,181 @@
...
+    tx20_wind_speed_kmh = float(tx20_sc) * 0.36;
+    tx20_wind_direction = tx20_sb;
+    std::string direction = DIRECTIONS.at(tx20_wind_direction);
+    ESP_LOGV(TAG, "WindSpeed %f, WindDirection %d, WindDirection Text %s", tx20_wind_speed_kmh, tx20_wind_direction,
+             direction.c_str());
+    this->wind_direction_sensor_->publish_state(tx20_wind_direction);
+    this->wind_speed_sensor_->publish_state(tx20_wind_speed_kmh);
+    this->wind_direction_text_sensor_->publish_state(direction);
+  } else {
+    ESP_LOGV(TAG, "Incorrect checksum!");
```
OttoWinter: Let's make this a warning

src/esphomelib/sensor/tx20_component.cpp
```diff
@@ -0,0 +1,181 @@
...
+  }
+}
+
+}  // namespace sensor
+
+ESPHOMELIB_NAMESPACE_END
+
+#endif  // USE_TX20
```
OttoWinter: Missing newline at end of file

OttoWinter (commented):
Great work on making this async 👍 

For the text sensor, please see comment in esphomeyaml - I'd prefer if the text sensor part be split off for now. It's too much of an architecture abuse (and I have the feeling it will bite me in the future - especially with changes to how esphomeyaml will conditionally compile files in the future to improve compilation speed)

src/esphomelib/sensor/tx20_component.cpp
```diff
@@ -0,0 +1,172 @@
...
+ESPHOMELIB_NAMESPACE_BEGIN
+
+namespace sensor {  // namespace
+
+static const char *TAG = "sensor.tx20";
+static const uint8_t MAX_BUFFER_SIZE = 41;
+static const uint16_t TX20_MAX_TIME = MAX_BUFFER_SIZE * 1220 + 5000;
+static const uint16_t TX20_BIT_TIME = 1200;
+static const std::string DIRECTIONS[] = {"N", "NNE", "NE", "ENE", "E", "ESE", "SE", "SSE",
+                                         "S", "SSW", "SW", "WSW", "W", "WNW", "NW", "NNW"};
```
OttoWinter: Remove `DIRECTIONS` now that there's no text sensor. Also `std::string` has the same issue as `std::vector<>`, it prevents compiler optimizations - it would have needed to be `static const char *DIRECTIONS[]`

src/esphomelib/sensor/tx20_component.cpp
```diff
@@ -0,0 +1,172 @@
...
+
+void TX20Component::setup() {
+  ESP_LOGCONFIG(TAG, "Setting up TX20...");
+  this->pin_->setup();
+  buffer_ = new uint16_t[MAX_BUFFER_SIZE];
+  attachInterrupt(this->pin_->get_pin(), TX20Component::pin_change_, CHANGE);
+}
+void TX20Component::dump_config() {
+  ESP_LOGCONFIG(TAG, "TX20:");
+  LOG_PIN(" Pin:", this->pin_);
```
OttoWinter:
```suggestion
  LOG_PIN("  Pin:", this->pin_);
```

src/esphomelib/sensor/tx20_component.h
```diff
@@ -0,0 +1,54 @@
...
+  TX20WindDirectionDegreesSensor *get_wind_direction_degrees_sensor() const;
+
+  void setup() override;
+  void dump_config() override;
+  float get_setup_priority() const override;
+  void loop() override;
+
+ protected:
+  static void pin_change_();
+  void decodeAndPublish_();
```
OttoWinter: `lower_snake_case`

src/esphomelib/sensor/tx20_component.h
```diff
@@ -0,0 +1,54 @@
+#include "esphomelib/defines.h"
+
+#ifdef USE_TX20
+
+#include "esphomelib/sensor/sensor.h"
+#include "esphomelib/text_sensor/text_sensor.h"
```
OttoWinter: Remove this

src/esphomelib/defines.h
```diff
@@ -612,4 +613,13 @@
   #endif
 #endif
 
+#ifdef USE_TX20
+  #ifndef USE_SENSOR
+    #define USE_SENSOR
+  #endif
+  #ifndef USE_TEXT_SENSOR
```
OttoWinter: Remove this too

OttoWinter (commented):
Ok, looking quite good now.

src/esphomelib/sensor/tx20_component.cpp
TheEggi: image

Was testing this for half a day - seems to be not good enough. Readded the check again.

src/esphomelib/sensor/tx20_component.cpp
TheEggi: Happens quite a few times -> Added a comment and will keep it as verbose.

src/esphomelib/sensor/tx20_component.cpp
```diff
@@ -0,0 +1,173 @@
...
+    this->wind_direction_degrees_sensor_->publish_state(float(tx20_wind_direction) * 22.5f);
+    this->wind_speed_sensor_->publish_state(tx20_wind_speed_kmh);
+
+  } else {
+    // sensor seems to produces quite a few checksum errors -> only in verbose mode
+    ESP_LOGV(TAG, "Incorrect checksum!");
+  }
+}
+
+uint16_t *TX20Component::buffer_ = new uint16_t[MAX_BUFFER_SIZE];
```
OttoWinter: This needs to be allocated in the constructor. Otherwise it might be allocated even when the user isn't using the TX20.

src/esphomelib/sensor/tx20_component.cpp
```diff
@@ -0,0 +1,169 @@
...
+    : Component(),
+      wind_speed_sensor_(new TX20WindSpeedSensor(wind_speed_name)),
+      wind_direction_degrees_sensor_(new TX20WindDirectionDegreesSensor(wind_direction_degrees_name)),
+      pin_(pin) {
+
+}
+
+void TX20Component::setup() {
+  ESP_LOGCONFIG(TAG, "Setting up TX20...");
+  TX20Component::buffer_ = new uint16_t[MAX_BUFFER_SIZE];
```
OttoWinter: buffer_ should be allocated in setup (or constructor). If it is allocated globally every user will lose quite a bit of RAM even if they don't use this integration (plus the allocator on the ESPs is sometimes a bit flakey when it is activated pre user code)

src/esphomelib/sensor/tx20_component.cpp
```diff
@@ -0,0 +1,169 @@
...
+
+void ICACHE_RAM_ATTR TX20Component::pin_change_() {
+  uint32_t now = micros();
+  if (!start_time_) {
+    buffer_[buffer_index_] = 1;
+    start_time_ = now;
+    buffer_index_++;
+    return;
+  }
+  uint32_t delay = now - start_time_;
```
OttoWinter: Updated this to simplify logic. Because now and start_time_ are unsigned integers, the result will always be positive and the correct number too.

OttoWinter (commented):
Okay, let's get this thing done! :)

I updated the PR a bit now (and left some comments for the changes). I will look later if it compiles correctly and if you agree with the changes I made I would be happy to merge this PR

src/esphomelib/sensor/tx20_component.cpp
TheEggi: Isn't there still the possibility of an overflow?

src/esphomelib/sensor/tx20_component.cpp
TheEggi: If this is the case it makes sense. I even had the code like that, but was under the impression that the code inside of the USE_TX20 only takes effect when it is in use and therefore moved it down to save a line.

kevank:
Is this complete?  Can we use a tx20?

TheEggi:
Currently you have to use the branch, but other than that it should be working. Update of the docs will follow on the weekend.

OttoWinter:
This moved [here](https://github.com/esphome/esphome/pull/275) (esphome-core is no longer used)

PR #338, opened 2018-12-28
Dump native API Server config on boot

voicevon:


src/esphomelib/api/api_server.cpp
```diff
@@ -54,7 +54,7 @@ void APIServer::loop() {
   }
 }
 void APIServer::dump_config() {
-
+  ESP_LOGI(TAG,"API Server:");
```
OttoWinter: This is not an info message, it's a config message.
Space missing after comma.

voicevon: sure, I will use ESP_LOGCONFIG().Thanks.

PR #339, opened 2018-12-28
How many components

voicevon:
This helped me to find out mqtt_message_trigger is not registered.  Perhaps need a better print style.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 .pioenvs
 .piolibdeps
+.vscode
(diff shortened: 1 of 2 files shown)


OttoWinter (commented):
No. This is
* not an info message - if each user would see this with normal debug level it's wrong.
* not even a debug message - maybe a very verbose message at most
* The log message is wrong. It may have been good fir debugging to have all the dashes in there but in most cases peoppe won't care about it.

PR #438, opened 2019-01-22
Add logging to NeoPixelBus

badbadc0ffee:
## Description:

Has anyone any (better) idea how to set up a local 'TAG' symbol in these template header-only implementations?

## Checklist:

### DIFF
--- a/src/esphomelib/light/neo_pixel_bus_light_output.h
+++ b/src/esphomelib/light/neo_pixel_bus_light_output.h
@@ -100,7 +100,6 @@ class NeoPixelBusLightOutputBase : public Component, public AddressableLight {
   bool has_requested_high_power_{false};
 #endif
 
-private:
   static const char* TAG;
 };
 
--- a/src/esphomelib/light/neo_pixel_bus_light_output.h
+++ b/src/esphomelib/light/neo_pixel_bus_light_output.h
@@ -79,6 +79,8 @@ class NeoPixelBusLightOutputBase : public Component, public AddressableLight {
 
   void setup() override;
 
+  void dump_config() override;
+
   void loop() override;
 
   float get_setup_priority() const override;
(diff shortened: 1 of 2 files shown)


OttoWinter (commented):
Previously, I've done this: Add `extern const char *NEO_PIXEL_BUS_TAG;` in .tcc and `const char *NEO_PIXEL_BUS_TAG = "...";` in .cpp

That may have the same result though as using the anonymous namespace. If you can, could you double-check the string is not included in the final elf with `strings firmware.elf` when neopixelbus is not used?

badbadc0ffee:
The anonymous namespace would have worked just as good: the string is not in the firmware binary when the neo_pixel_bus_light ist not configured.
But finally, I like this variant better because the same symbol TAG can be used in the logging macros as everywhere else.

src/esphomelib/light/neo_pixel_bus_light_output.h
```diff
@@ -97,6 +99,9 @@ class NeoPixelBusLightOutputBase : public Component, public AddressableLight {
   PowerSupplyComponent *power_supply_{nullptr};
   bool has_requested_high_power_{false};
 #endif
+
+private:
```
OttoWinter:
```suggestion
 protected:
```

a) indent with 1 space as google guidelines say
b) use protected (long story short, I prefer protected for various reasons concerning extensibility)

badbadc0ffee: I didn't realize that there is this "suggested change" feature now. Will use it next time...

OttoWinter: No no, suggested change is only there to make it easier _for you_. I don't care if you use it or not :)

OttoWinter (approved):

src/esphomelib/light/neo_pixel_bus_light_output.h
badbadc0ffee: Sure! But I think I like it...

PR #483, opened 2019-02-10
Enable use of alternate hardware UARTs for logging

brandond:
## Description:
Enable use of Serial1 on ESP8266 and Serial1/Serial2 on ESP32 for logging.
This is frequently done on ESP8266 to allow use of Serial for UART TX+RX,
while  maintaining logging output on Serial1 which is TX-only via GPIO2.

**Pull request in [esphomeyaml](https://github.com/OttoWinter/esphomeyaml) with YAML changes (if applicable):** OttoWinter/esphomeyaml#427
**Pull request in [esphomedocs](https://github.com/OttoWinter/esphomedocs) with documentation (if applicable):** OttoWinter/esphomedocs#161

## Checklist:
  - [X] The code change is tested and works locally.

  - [x] Documentation added/updated in [esphomedocs](https://github.com/OttoWinter/esphomedocs).

### DIFF
--- a/src/esphomelib/log_component.h
+++ b/src/esphomelib/log_component.h
@@ -17,6 +17,22 @@
 
 ESPHOMELIB_NAMESPACE_BEGIN
 
+/** Enum for logging UART selection
+ *
+ * Advanced configuration (pin selection, etc) is not supported.
+ */
+enum UARTSelection {
+  ESPHOMELIB_UART0 = 0,
+  ESPHOMELIB_UART1,
+#ifdef ARDUINO_ARCH_ESP32
+  ESPHOMELIB_UART2
+#endif
+#ifdef ARDUINO_ARCH_ESP8266
+  ESPHOMELIB_UART0_SWAP
+#endif
+};
+
+
 /** A simple component that enables logging to Serial via the ESP_LOG* macros.
  *
  * This component should optimally be setup very early because only after its setup log messages are actually sent.
@@ -29,13 +45,16 @@ class LogComponent : public Component {
    * @param baud_rate The baud_rate for the serial interface. 0 to disable UART logging.
    * @param tx_buffer_size The buffer size (in bytes) used for constructing log messages.
    */
-  explicit LogComponent(uint32_t baud_rate = 11520, size_t tx_buffer_size = 512);
+  explicit LogComponent(uint32_t baud_rate = 115200, size_t tx_buffer_size = 512, UARTSelection uart = ESPHOMELIB_UART0);
 
   /// Manually set the baud rate for serial, set to 0 to disable.
   void set_baud_rate(uint32_t baud_rate);
 
   /// Set the buffer size that's used for constructing log messages. Log messages longer than this will be truncated.
   void set_tx_buffer_size(size_t tx_buffer_size);
+
+  /// Get the UART used by the logger.
+  UARTSelection get_uart() const;
 
   /// Set the global log level. Note: Use the ESPHOMELIB_LOG_LEVEL define to also remove the logs from the build.
   void set_global_log_level(int log_level);
@@ -68,6 +87,8 @@ class LogComponent : public Component {
  protected:
   uint32_t baud_rate_;
   std::vector<char> tx_buffer_;
+  UARTSelection uart_{ESPHOMELIB_UART0};
+  HardwareSerial *hw_serial_{nullptr};
   int global_log_level_{ESPHOMELIB_LOG_LEVEL};
   struct LogLevelOverride {
     std::string tag;
--- a/src/esphomelib/log_component.h
+++ b/src/esphomelib/log_component.h
@@ -22,13 +22,13 @@ ESPHOMELIB_NAMESPACE_BEGIN
  * Advanced configuration (pin selection, etc) is not supported.
  */
 enum UARTSelection {
-  ESPHOMELIB_UART0 = 0,
-  ESPHOMELIB_UART1,
+  UART_SELECTION_UART0 = 0,
+  UART_SELECTION_UART1,
 #ifdef ARDUINO_ARCH_ESP32
-  ESPHOMELIB_UART2
+  UART_SELECTION_UART2
 #endif
 #ifdef ARDUINO_ARCH_ESP8266
-  ESPHOMELIB_UART0_SWAP
+  UART_SELECTION_UART0_SWAP
 #endif
 };
 
@@ -45,7 +45,7 @@ class LogComponent : public Component {
    * @param baud_rate The baud_rate for the serial interface. 0 to disable UART logging.
    * @param tx_buffer_size The buffer size (in bytes) used for constructing log messages.
    */
-  explicit LogComponent(uint32_t baud_rate = 115200, size_t tx_buffer_size = 512, UARTSelection uart = ESPHOMELIB_UART0);
+  explicit LogComponent(uint32_t baud_rate = 115200, size_t tx_buffer_size = 512, UARTSelection uart = UART_SELECTION_UART0);
 
   /// Manually set the baud rate for serial, set to 0 to disable.
   void set_baud_rate(uint32_t baud_rate);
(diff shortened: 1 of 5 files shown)


src/esphomelib/uart_component.cpp
```diff
@@ -22,15 +22,20 @@ UARTComponent::UARTComponent(int8_t tx_pin, int8_t rx_pin, uint32_t baud_rate)
...
+  } elif (this->rx_pin_ == 16 && this->tx_pin_ == 17) {
+    this->hw_serial_ = &Serial2
   } else {
     this->hw_serial_ = new HardwareSerial(next_uart_num++);
   }
 #endif
 #ifdef ARDUINO_ARCH_ESP8266
   if (this->rx_pin_ == 3 && this->tx_pin_ == 1) {
     this->hw_serial_ = &Serial;
+  } elif (this->rx_pin_ == 8 && this->tx_pin_ == 2) {
```
OttoWinter: It looks like the ESP8266 does not support RX on the UART1 interface (see https://github.com/esp8266/Arduino/blob/master/cores/esp8266/uart.cpp#L631-L637).

OttoWinter: `elif` -> `else if`

OttoWinter (commented):
While reviewing this PR I realized some of the logic in the uart component was a bit funky :)

For example, the HW serial should be used to if either one of RX or TX pin matches (or both). Until now the HW serial was only used on the ESP8266 iff tx&rx were used.

Created https://github.com/esphome/esphome-core/pull/487 to fix this. I think this PR can be rebased on that once that's complete.

src/esphomelib/uart_component.cpp
```diff
@@ -22,15 +22,20 @@ UARTComponent::UARTComponent(int8_t tx_pin, int8_t rx_pin, uint32_t baud_rate)
     : tx_pin_(tx_pin), rx_pin_(rx_pin), baud_rate_(baud_rate) {
 #ifdef ARDUINO_ARCH_ESP32
   if (this->rx_pin_ == 3 && this->tx_pin_ == 1) {
-    // Default UART for logging
     this->hw_serial_ = &Serial;
+  } elif (this->rx_pin_ == 9 && this->tx_pin_ == 10) {
+    this->hw_serial_ = &Serial1
+  } elif (this->rx_pin_ == 16 && this->tx_pin_ == 17) {
+    this->hw_serial_ = &Serial2
```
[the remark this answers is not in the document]
brandond: Sure, but UART1 and UART2 are set up on these pins by default by the Arduino framework. I'm not sure what happens if someone tries to re-use those same pins on another UART - I guess nothing, as long as they don't also try to use the Arduino global Serial instances that are preconfigured to use those pins?

src/esphomelib/uart_component.cpp
brandond: UART1 supports RX, it's just that the pin (IO8) is also used for SDIO_DATA1 and is disabled when SPI is in use - which is basically always.

UART0 is also interesting in that it supports RTS/CTS on IO13/IO12, and RTS/CTS and RX/TX can be swapped. It's not anywhere near the flexibility of ESP32, but it's not bad given the overall low pinout of the device.

brandond:
I just rebased on top of your fixes. I opted to keep the Serial1/Serial2 reuse stuff since I think it might be useful. I'll put in a docs PR to describe how to make use of it.

brandond:
@OttoWinter can you sanity-check my approach here? I'd kind of like to keep it simple and allow users to move the logger over to another hardware UART. However, since there's a full UART component now, it feels like the logger should really use that instead of just using the raw global Serial object. We'd need to expose some of the low-level hardware capabilities (`setDebugOutput`, `swap`, etc), through UART, but that's not terrible. 

That would change the logger config to something like this:
```
uart:
  - tx_pin: 1
    rx_pin: 3
    baud_rate: 9600
    id: uart0
  - tx_pin: 2
    rx_pin: 8
    baud_rate: 115200
    id: uart1

logger:
  level: DEBUG
  id: logger
  #baud_rate: 115200 # Exclusive of uart_id
  uart_id: uart1
```

This would also mean that UARTs need to be set up before the logger; I'm not sure if that will break anything else?

OttoWinter:
Yeah that would be an option. On the ESP32 that would add any functionality. On the ESP8266 it would enable the use of software serial for sending.

However, I don't think it's worth the effort TBH. As you said, the UART component would need to be set up before the logger, and a whole bunch of edge cases would need to be handled.

Then there's the problem of speed. The logger is called often and should not do much work, software serial on ESP8266 would probably make any log call quite slow. Plus the logger should never break - the logger is the first thing that's initialized, if that breaks the user won't:
* get any logs showing what went wrong
* even get OTA safe mode.

And then there's the question why we should do this. Sure it would maybe be cleaner on a high level with more of a plug-and-play nature, but I doubt any user would ever use the added functionality that offers.

brandond:
I really just need to get the logger off to UART1 on esp8266 so I can free up UART0 for the communication with the coprocessor on #306 while not loosing the serial log capability. This is what the stock Tuya firmware does.

brandond:
OK, I think this is a fairly simple approach that lets me accomplish what I'm trying to do without complicating things too much.

OttoWinter (approved):
👍

src/esphomelib/log_component.h
```diff
@@ -17,6 +17,22 @@
 
 ESPHOMELIB_NAMESPACE_BEGIN
 
+/** Enum for logging UART selection
+ *
+ * Advanced configuration (pin selection, etc) is not supported.
+ */
+enum UARTSelection {
+  ESPHOMELIB_UART0 = 0,
```
OttoWinter: For enums, I prefer to prefix them with the enum name (I think this is google style too, but not sure).

So `UART_SELECTION_UART0 = 0,`

(Or in this case mostly just so that I don't have to change this to `ESPHOME_UART0` in https://github.com/esphome/esphome-core/pull/494 again 😜 )

brandond: Can do! I think I copied it from the log stuff - ESPHOMELIB_LOG_X etc.

brandond:
I finally got a chance to test this on real hardware, and it mostly looks good. There's some Arduino stuff that's hardcoded to output on a specific port using the DEBUG_ESP_PORT macro; I'll have to add a note to the docs about that.

src/esphomelib/uart_component.cpp
```diff
@@ -115,11 +120,22 @@ void UARTComponent::flush() {
...
   ESP_LOGCONFIG(TAG, "Setting up UART bus...");
+  // Use Arduino HardwareSerial UARTs if all used pins match the ones
+  // preconfigured by the platform. For example if RX disabled but TX pin
+  // is 1 we still want to use Serial.
   if (this->tx_pin_.value_or(1) == 1 && this->rx_pin_.value_or(3) == 3) {
-    // use UART0 if all used pins match the ones on the UART bus
-    // for example if RX disabled but TX pin is 1 we still want to use Serial.
     this->hw_serial_ = &Serial;
     this->hw_serial_->begin(this->baud_rate_);
+    this->hw_serial_->setDebugOutput(false);
```
OttoWinter: I do not think we should disable debug output here. If the user doesn't want debug output on a serial port they should disable the logger component for it. 

If they don't, even though we disable the SDKs debug output here, the user will still get debug output through esphome's own logging engine.

brandond: I'll re-test without this but with the platformio build flag. My experience was that I still got debug output on UART0 unless I called setDebugOutput(false) on it here. This is due to the Arduino framework enabling it by default even before the esphome logger takes over. The esphome logger calls setDebugOutput(true) on the uart selected for logging, so the change is just to put debugging on the logger UART, not to disable it entirely.

brandond: I guess that in the case of UART and Logger sharing a port, the UART would disable it after the Logger enabled it, so you're right - this isn't safe to do unconditionally.

OttoWinter (approved):
Ok, this looks great now! Thanks!

PR #501, opened 2019-02-15
Add light partition platform

### LINKED ISSUE #44: Segments for NeoPixels
**Is your feature request related to a problem/use-case? Please describe.**
Using a strip of Neopixels it may be desirable to create affects or control segments of the strip as individual lights rather than as a whole. Whilst access to pixels via the "- addressable_lambda:" option is possible its not as user friendly or flexible as accessing the strip as a whole

**Describe the solution you'd like:**
Separate neopixel functionality into a hub that defines the Pin and neopixel parameters (method/type etc.),  and then allow the definition of a neopixel light to reference the hub, start pixel and number of pixels. The full range of effects would then be available to that segment independent of other segments within the same neopixel string (pin)

**Additional context:**
The  ws2812fx library implements segments for this feature.

OttoWinter:
## Description:

**Related issue (if applicable):** fixes https://github.com/esphome/feature-requests/issues/44

## Checklist:

### DIFF
--- a/src/esphome/light/addressable_light.cpp
+++ b/src/esphome/light/addressable_light.cpp
@@ -131,6 +131,69 @@ bool AddressableLight::is_effect_active() const {
 void AddressableLight::set_effect_active(bool effect_active) {
   this->effect_active_ = effect_active;
 }
+void AddressableLight::write_state(LightState *state) {
+  this->write_state(state, 0, this->size());
+}
+
+int32_t PartitionLightOutput::size() const {
+  auto last_seg = this->segments_[this->segments_.size() - 1];
+  return last_seg.get_dst_offset() + last_seg.get_size();
+}
+ESPColorView PartitionLightOutput::operator[](int32_t index) const {
+  uint32_t lo = 0;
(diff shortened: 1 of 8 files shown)


src/esphome/light/addressable_light.cpp
```diff
@@ -131,6 +131,69 @@ bool AddressableLight::is_effect_active() const {
...
+}
+void PartitionLightOutput::write_state(LightState *state, int32_t begin, int32_t end) {
+  for (auto seg : this->segments_) {
+    int32_t dbegin = seg.get_dst_offset();
+    if (begin > dbegin)
+      continue;
+
+    int32_t sbegin = seg.get_src_offset();
+    int32_t send = begin + std::min(seg.get_size(), end - dbegin);
+    seg.get_src()->write_state(state, sbegin, send);
```
OttoWinter: This is not 100% correct (light correction, brightness applies to entire strip), but I think for most setups this won't matter. Because it's hard to fix I will defer that to the future (if it ever comes up)

PR #538, opened 2019-03-01
Add SDS011

puuu:
## Description:

**Related issue (if applicable):** 

**Pull request in [esphome](https://github.com/esphome/esphome) with python changes (if applicable):** esphome/esphome#467
**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#185

## Checklist:

  - [x] Documentation added/updated in [esphome-docs](https://github.com/esphome/esphome-docs).


src/esphome/sensor/sds011_component.cpp
```diff
@@ -0,0 +1,258 @@
...
+void SDS011Component::parse_data_() {
+  this->status_clear_warning();
+  const float pm_2_5_concentration = this->get_16_bit_uint_(2) / 10.0f;
+  const float pm_10_0_concentration = this->get_16_bit_uint_(4) / 10.0f;
+  const uint16_t device_id = this->get_16_bit_uint_(6);
+
+  ESP_LOGD(TAG,
+           "Got PM2.5 Concentration: %.1f µg/m³, PM10.0 Concentration: %.1f "
+           "µg/m³ from device 0x%04x",
+           pm_2_5_concentration, pm_10_0_concentration, device_id);
```
OttoWinter: Can there be more than one device connected to one bus? If so, we should split add a device_id option so that users can set up multiple instances of this class and select which sensor to read from.

src/esphome/sensor/sds011_component.cpp
```diff
@@ -0,0 +1,258 @@
...
+  ESP_LOGCONFIG(TAG, "  Query mode: %u", this->query_mode_);
+  LOG_SENSOR("  ", "PM2.5", this->pm_2_5_sensor_);
+  LOG_SENSOR("  ", "PM10.0", this->pm_10_0_sensor_);
+}
+
+void SDS011Component::loop() {
+  const uint32_t now = millis();
+  if ((now - this->last_transmission_ >= 500) && this->data_index_) {
+    // last transmission too long ago. Reset RX index.
+    ESP_LOGD(TAG, "last transmission too long ago. Reset RX index.");
```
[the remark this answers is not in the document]
puuu: I changed this to verbose.

src/esphome/sensor/sds011_component.cpp
```diff
@@ -0,0 +1,258 @@
...
+  while (this->available() != 0) {
+    this->read_byte(&this->data_[this->data_index_]);
+    auto check = this->check_byte_();
+    if (!check.has_value()) {
+      // finished
+      this->parse_data_();
+      this->data_index_ = 0;
+    } else if (!*check) {
+      // wrong data
+      ESP_LOGD(TAG, "Wrong data.");
```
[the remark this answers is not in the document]
puuu: I changed this to verbose and improved the message.

src/esphome/sensor/sds011_component.cpp
```diff
@@ -0,0 +1,258 @@
...
+  command_data[1] = SDS011_SET_MODE;
+  command_data[2] = this->query_mode_ ? SDS011_MODE_REPORT_QUERY : SDS011_MODE_REPORT_ACTIVE;
+  command_data[13] = 0xff;
+  command_data[14] = 0xff;
+  this->sds011_write_command_(command_data);
+
+  uint8_t update_interval_minutes = this->get_update_interval() / 1000.0f / 60.0f;
+  this->set_update_interval((uint32_t)update_interval_minutes * 1000 * 60);
+  if (update_interval_minutes > 30) {
+    ESP_LOGW(TAG, "Update interval is longer than 30 min.");
```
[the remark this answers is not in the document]
puuu: According to the [protocol description](https://nettigo.pl/attachments/415), the interval should be between 1-30 minutes. I did not tested what will happen with higher values.

src/esphome/sensor/sds011_component.cpp
```diff
@@ -0,0 +1,258 @@
...
+  uint8_t command_data[SDS011_DATA_REQUEST_LENGTH] = {0};
+  command_data[0] = SDS011_COMMAND_REPORT_MODE;
+  command_data[1] = SDS011_SET_MODE;
+  command_data[2] = this->query_mode_ ? SDS011_MODE_REPORT_QUERY : SDS011_MODE_REPORT_ACTIVE;
+  command_data[13] = 0xff;
+  command_data[14] = 0xff;
+  this->sds011_write_command_(command_data);
+
+  uint8_t update_interval_minutes = this->get_update_interval() / 1000.0f / 60.0f;
+  this->set_update_interval((uint32_t)update_interval_minutes * 1000 * 60);
```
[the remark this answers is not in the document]
puuu: The working period should be between 1-30 minutes and is given in minutes. (0 means continuous measurements). `update_interval_minutes` is sent to the sensor.  `update_interval` is only used to poll the sensor in query mode. . Casting in Python would be alright, but I do not know how to restrict the time type (in interval and maximum). Do you have an example?

src/esphome/sensor/sds011_component.cpp
puuu: It is not directly mentioned that multiple sensors are possible. However, since the data frames in the protocol contain a device id and a broadcast id (`0xffff`) exists, one would think that this is possible. But for me is not clear how possible collision will be handled on the bus. I have only one device and can not test it.

IMHO, to support multiple sensors on the bus, we need to split the component into two parts (similar to the dallas component):
1. Hub component to configure the uart port and handle incoming frames.
2. Sensor component with device_id (default to broadcast id) .

I had this in mind (and basically like that idea), but thought it was too over-engineered, especially since it is not clear if multiple sensors are possible.

src/esphome/sensor/sds011_component.cpp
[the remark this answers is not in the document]
OttoWinter: Ok, it's rather simple (I'll add it to the esphome pr directly). We should remove the code here, components shouldn't set their own properties - it probably works now but it could cause some trouble later.

src/esphome/sensor/sds011_component.cpp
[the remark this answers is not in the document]
OttoWinter: Ok then, that's ok.

But we should move this warning then to `dump_config` - setup is called very early on and unless the user is connected via USB they won't see the warning.

src/esphome/sensor/sds011_component.cpp
OttoWinter: Ok, I guess especially since we support SW UART on the ESP8266s there's not a real reason to support more than one device per bus. Even if we could, the sensors could run into trouble when they both hold the bus high/low at the same time.

OttoWinter: Let's just remove that from the logs then, since the normal user shouldn't be concerned with this.

src/esphome/sensor/sds011_component.cpp
```diff
@@ -0,0 +1,243 @@
...
+  command_data[1] = SDS011_SET_MODE;
+  command_data[2] = update_interval_minutes;
+  command_data[13] = 0xff;
+  command_data[14] = 0xff;
+  this->sds011_write_command_(command_data);
+
+  if (this->get_query_mode()) {
+    this->update();
+    this->set_timeout("delayed-update", 35000, [this]() {
+      this->set_interval("update", this->get_update_interval(), [this]() { this->update(); });
```
OttoWinter: We should not override the PollingComponent's update `set_interval`.

What is the reason for this delayed second update?

src/esphome/sensor/sds011_component.cpp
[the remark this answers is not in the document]
puuu: Ok, I moved it to `dump_config()` and added an additional check in python.

src/esphome/sensor/sds011_component.cpp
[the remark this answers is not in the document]
puuu: Thank you! I removed the casting.

src/esphome/sensor/sds011_component.cpp
puuu: Done

src/esphome/sensor/sds011_component.cpp
puuu: > We should not override the PollingComponent's update set_interval.

What do you mean with override? This component does not inherit from the `PollingComponent` class. However, with `query_mode`, it mimics the `PollingComponent` class. If you don't like it, I will change it.

> What is the reason for this delayed second update?

The query command only ask for the last measurement value and do not force a measurement. For working periods (update_interval >= 1min), the sensor need 30s for the measurement (mainly to get fresh air from the fan). So for a power cycle of esp and sensor, we have to wait 30s to get a valid value. In the case, the esp rebooted or was in deepsleep (esp8266), we can read a valid value with the first `update()`.

OttoWinter: > What do you mean with override? This component does not inherit from the PollingComponent class. However, with query_mode, it mimics the PollingComponent class. If you don't like it, I will change it.

Oh, I assumed it was overriding `PollingComponent`. In that case we should override `PollingComponent` and `return` in update if we're not in query mode.

> The query command only ask for the last measurement value and do not force a measurement. For working periods (update_interval >= 1min), the sensor need 30s for the measurement (mainly to get fresh air from the fan). So for a power cycle of esp and sensor, we have to wait 30s to get a valid value. In the case, the esp rebooted or was in deepsleep (esp8266), we can read a valid value with the first update().

If the user only sees this on the first time they plug the device in I think it's not a problem. Deep sleep is likely also not that important here because the sensor surely uses tons of power that would quickly drain all batteries.

OttoWinter (approved):
Looks good now, change `Component` to `PollingComponent` and it's ready to merge.

src/esphome/sensor/sds011_component.cpp
puuu: As far as I can see, when we switch to `PollingComponent`, we can not delay the first query and always get the value of the previous measurement (which is several minutes old). IMHO, `query_mode` only makes sense for deep_sleep

OttoWinter (approved):
LGTM! Thanks :+1:

src/esphome/sensor/sds011_component.cpp
OttoWinter: That is true, however:

- `deep_sleep` does not really make sense with this integration I think. I'm pretty sure the sensor uses tons of energy even when in idle mode (might be wrong about this). And when it's in non-idle mode it surely uses a ton of energy too - so you'd only get at most a week of deep sleep uptime
- While delaying the initial reading might be the correct way of doing it, I think it's not that important, because: It only ever happens when the user power-cycles the setup or on initial data - IMO those two cases happen so seldomly that it's ok if we do not handle those correctly

OttoWinter:
Ok, so you were right about not using PollingComponent - it just doesn't fit very well here.

I realized that the update interval is only in minutes anyway with this sensor (since query mode was removed). So I think it is now cleaner to just have a `update_interval_min_` property that denotes the update interval in minutes. Then the python part also gets a bit cleaner.

The effect of this is also that to enable the continuous-measurement mode, the user needs to type `update_interval: 0min`, but I think that's ok (all values from 0s to 59s were essentially the same, so having one value is cleaner)

If you're ok with these changes, I'd be happy to merge. Also, please tell me if you prefer me not touching your git branches, I just think this way the whole conversation part and potential misunderstandings is removed - code talks :D

puuu:
OK. Tested and works with my last commit in the python part.

Thank you.
Fell free to merge!

PR #548, opened 2019-03-12
GroupSensorComponent group binary_sensors to one Sensor.

mvturnho:
## Description:
Class to group binary_sensors to one Sensor.
You add binary_sensors to this sensor with a specific value.
Each binary sensor represents a float value in the group.

When the binary_sensor reports a true value for its state, then the float value it represents is added to the total_current_value. Only when the  total_current_value changed and at least one sensor reports an active state we publish the sensors average value.

## Checklist:


src/esphome/defines.h
```diff
@@ -230,6 +231,15 @@
...
 #endif
 
+#ifdef USE_GROUP_SENSOR
+#ifndef USE_SENSOR
+#define USE_SENSOR
+#endif
+#ifndef USE_BINARY_SENSOR
+#define USE_BINARY_SENSOR
+#endif
+#endif
```
OttoWinter: Remove this, these build flags are set automatically in python.

src/esphome/sensor/group_sensor.cpp
```diff
@@ -0,0 +1,73 @@
...
+}
+
+/**
+ * The loop checks all binary_sesnor states
+ * When the binary_sensor reports a true value for its state, then the float value it represents is added to the
+ * total_current_value
+ *
+ * Only when the  total_current_value changed and at least one sensor reports an active state we publish the sensors
+ * average value
+ * */
```
OttoWinter: Please place the documentation comments in the header file.

src/esphome/sensor/group_sensor.cpp
```diff
@@ -0,0 +1,73 @@
...
+      num_active_sensors++;
+      total_current_value += bs->value;
+    }
+  }
+  // check it at least one binary_sensor is active and the value changed.
+  if ((last_value_ != total_current_value) && num_active_sensors > 0) {
+    ESP_LOGD(TAG, "%s - value: %f", this->name_.c_str(), total_current_value / num_active_sensors);
+    // publish the average value
+    this->publish_state(total_current_value / num_active_sensors);
+  }
```
OttoWinter: Should we publish something when no channel is active? I'm thinking it could be useful to know when nothing is pressed. Maybe something like `this->publish_state(NAN)` (used in a couple of other places too).

mvturnho: I think the mqtt messages will overflow becouse it wil the publish `NAN ` with every iteration of the loop. I tested this (by accident). Now the activation of a binary sensor is the trigger to publish its value. An alternative could be that when releasing all buttons we publish a `NAN `once.

Maybe we should make this behaviour a setting in the yaml. Like 
`trigger_sensor: press|release|pressrelease`

mvturnho: GROUP sensor type works like this now:
- when touched: publish active channel value once until release or active channels change.
- when release: publish nan value to indicate the release.

src/esphome/sensor/binary_sensor_map.cpp
```diff
@@ -0,0 +1,96 @@
...
+    if (this->last_touched_) {
+      this->last_touched_ = false;
+      ESP_LOGD(TAG, "%s - release value: nan", this->name_.c_str());
+      this->publish_state(NAN);
+    }
+  }
+}
+
+void BinarySensorMap::process_slider_() { ESP_LOGD(TAG, "SLIDER is not implemented yet."); }
+void BinarySensorMap::process_wheel_() { ESP_LOGD(TAG, "WHEEL is not implemented yet."); }
```
OttoWinter: If it is not implemented yet we shouldn't have code for it yet. We can still add it later when it is done. This will just unnecessarily take up space in flash.

src/esphome/sensor/binary_sensor_map.cpp
```diff
@@ -0,0 +1,96 @@
...
+static const char *TAG = "sensor.binary_sensor_map";
+
+BinarySensorMapChannel::BinarySensorMapChannel(binary_sensor::BinarySensor *sensor, float value) {
+  this->binary_sensor = sensor;
+  this->value = value;
+}
+
+BinarySensorMap::BinarySensorMap(const std::string &name) : Sensor(name) {}
+
+void BinarySensorMap::setup() { ESP_LOGCONFIG(TAG, "Setting up binary_sensor_map '%s'...", this->name_.c_str()); }
```
OttoWinter: If there's no setup logic we should remove this. The logs are only there for components that do some setup (and so it can quickly be found which component's setup failed)

src/esphome/sensor/binary_sensor_map.cpp
```diff
@@ -0,0 +1,96 @@
+#include "esphome/defines.h"
+
+#ifdef USE_BINARY_SENSOR_MAP
+
+#include "esphome/sensor/binary_sensor_map.h"
+
+#include "esphome/log.h"
+#include <bitset>
```
OttoWinter: This include is not used.

src/esphome/sensor/binary_sensor_map.h
```diff
@@ -0,0 +1,81 @@
...
+namespace sensor {
+
+enum {
+  BINARY_SENSOR_MAP_TYPE_GROUP = 0x00,
+  BINARY_SENSOR_MAP_TYPE_SLIDER = 0x01,
+  BINARY_SENSOR_MAP_TYPE_WHEEL = 0x02,
+};
+
+/** helper class to combine the binary_sensor and its float value to one objects. **/
+class BinarySensorMapChannel {
```
OttoWinter: If this is only going to be a data class, and has no logic itself then make it a `struct`. Otherwise make the attributes protected and write getters.

mvturnho:
@OttoWinter are there things left that need to be addressed for this PR?

src/esphome/sensor/binary_sensor_map.cpp
```diff
@@ -0,0 +1,82 @@
...
+    }
+    cur_sens++;
+  }
+  // check if the sensor map was touched
+  if (touched) {
+    // did the bit_mask change or is it a new sensor touch
+    if ((last_mask_ != mask) || !this->last_touched_) {
+      this->last_touched_ = true;
+      float publish_value = total_current_value / num_active_sensors;
+      ESP_LOGD(TAG, "%s - touched value: %f", this->name_.c_str(), publish_value);
```
OttoWinter: Use `%.3f` here instead of plain `%f`. Otherwise a lot of decimal places will be printed which makes the logs noisy

mvturnho:
Hi @OttoWinter can we merge this PR or does it need more changes?

OttoWinter:
Hi @mvturnho,

Sorry for the delay, I was working on some pretty big refactors in the esphome codebase (see esphome repo). The good thing is that it's almost done now and that new integrations will become easier to create in the future. Unfortunately that also means that the PR here will need to be update too.

I've planned to port some of the existing PRs over to the new API soon, but if you want you can already try it now. Sorry for not being able to merge this PR rn, but hopefully the changes will improve ESPHome and the contribution process a lot.

mvturnho:
Yes I noticed your work so I thought I wait with the PR until you finished. Then see if I can port this component to the new dev branch and create a new PR.

So no problem. I will give it a try myself.

OttoWinter:
Just a heads-up, the dev branch API is still a *bit* unstable (see cleanup-dashboard branch). Basically, now that we already have a breaking change and all PRs will be invalidated anyway, I'm trying to refactor a bunch of the core API to make creating integrations easier.

PR #558, opened 2019-03-21
cse7766 fix

brianrjones69:
## Description:
cse7766 fix.  

https://github.com/esphome/issues/issues/143

## Checklist:


brianrjones69:
This version works for my S31's.

src/esphome/sensor/cse7766.cpp
```diff
@@ -39,34 +39,32 @@ void CSE7766Component::loop() {
...
 bool CSE7766Component::check_byte_() {
   uint8_t index = this->raw_data_index_;
   uint8_t byte = this->raw_data_[index];
   if (index == 0) {
-    // Header, usually 0x55, contains data about calibration etc.
-    // this is validated in parse_data_
-    return true;
+    if ((byte != 0x55) && ((byte & 0xF0) != 0xF0) && (byte != 0xAA)) {
+      return false;
+    }
```
OttoWinter: There should be a `return true;` after the if statement.

src/esphome/sensor/cse7766.cpp
```diff
@@ -39,34 +39,32 @@ void CSE7766Component::loop() {
...
   }
 
   if (index == 1) {
     if (byte != 0x5A) {
       ESP_LOGV(TAG, "Invalid Header 2 Start: 0x%02X!", byte);
       return false;
     }
-
-    return true;
```
OttoWinter: Re-add this line please. This function is intentionally laid out in a way that for each index the function will execute up to a certain point, not until the end of the function.

src/esphome/sensor/cse7766.cpp
```diff
@@ -101,56 +95,79 @@ void CSE7766Component::parse_data_() {
...
+    this->power_sensor_->publish_state(power);
+
+  this->voltage_acc_ = 0.0f;
+  this->current_acc_ = 0.0f;
+  this->power_acc_ = 0.0f;
+  this->voltage_counts_ = 0;
+  this->power_counts_ = 0;
+  this->current_counts_ = 0;
 }
+void CSE7766Component::setup() {}
```
OttoWinter: If this is empty it should be removed.

src/esphome/sensor/cse7766.cpp
```diff
@@ -101,56 +95,79 @@ void CSE7766Component::parse_data_() {
...
-  float power = this->power_acc_ / d;
+  
+  float voltage = this->voltage_counts_ > 0 ? this->voltage_acc_ / this->voltage_counts_ : 0.0;
+  float current = this->current_counts_ > 0 ? this->current_acc_ / this->current_counts_ : 0.0;
+  float power = this->power_counts_ > 0 ? this->power_acc_ / this->power_counts_ : 0.0;
+
+  ESP_LOGD(TAG, "Got voltage_acc=%.2f current_acc=%.2f power_acc=%.2f", this->voltage_acc_, this->current_acc_, 
+           this->power_acc_);
+  ESP_LOGD(TAG, "Got voltage_counts=%d current_counts=%d power_counts=%d", this->voltage_counts_, this->current_counts_, 
+           this->power_counts_);
```
OttoWinter: The user doesn't care about the acc and count numbers, please make these verbose or remove them.

OttoWinter (commented):
Thanks! I'm wondering though, in what way is this code different from the one seen here: https://github.com/esphome/esphome-core/compare/fix-cse7766 (that apparently does not work...) - that branch contains more or less the same fixes

One difference I see is that voltage & current must be "ok" for a voltage reading to be published. Do you know if that is the cause?

src/esphome/sensor/cse7766.cpp
brianrjones69: I have to agree with you for the most part on your (new) version of the code.  In fact, the only problem I had was it was reading current (very small) when it should have been 0.

The spec sheet (see Figure 5, 6 and 7) says that bit3 and bit2 must be 0 for all readings.  

For power, bit1 must be 0, otherwise power is 0.
For current, if measured power is 0 then current is 0.

I saw that some other people are using the cse7759B.  The only datasheet I found for it was in Chinese, but the figures 5,6 and 7 look to be the same.  I didn't really have the time for a detailed analysis, but I will look later tonight and fix the items above.

Thanks for being patient with me here as this GitHub stuff is new to me.  Getting someone else's coding style right is another story.  Python though - I cant help you with!

Fix is on the way

brianrjones69:
I had a ton of issues with the lint to get the tests to pass.  I installed a tool so I could run it from within Visual Studio Code but it didn't like the .clang-format file in the project.  I then used the one from the master and it fixed some things but broke some others.  One setting it didn't like is:

`AlwaysBreakTemplateDeclarations: 'MultiLine'`

I am using Windows 10, so maybe I cant even get the version of tools needed, but do you have any docs on setting up the linter?

OttoWinter:
@brianrjones69 No there are no such instructions, mainly because it's *really* hard to install the two tools. That's why the CI tool prints a whole diff so that it can be copied and applied directly. the issue you're likely seeing is that you're not using the correct version of clang-format, version 7 is required.

src/esphome/sensor/cse7766.cpp
```diff
@@ -15,6 +15,7 @@ void CSE7766Component::loop() {
   const uint32_t now = millis();
   if (now - this->last_transmission_ >= 500) {
     // last transmission too long ago. Reset RX index.
+    ESP_LOGV(TAG, "%d milliseconds have passed since the last transmission.  Resetting index", now);
```
OttoWinter:
```suggestion
    ESP_LOGV(TAG, "%u milliseconds have passed since the last transmission.  Resetting index", now - this->last_transmission_);
```

Technically the log statement is incorrect. It can also be dropped though.

brianrjones69:
@OttoWinter I dont think that the timing code should really be there at all.  If we haven't seen anything in 500 milliseconds,  we probably never will see anything.  I think we should remove it completely.

src/esphome/sensor/cse7766.cpp
```diff
@@ -27,7 +27,6 @@ void CSE7766Component::loop() {
     if (!this->check_byte_()) {
       this->raw_data_index_ = 0;
       this->status_set_warning();
-      continue;
```
OttoWinter: Why was this continue removed? Without this, the code will accept any packet even if the checksum is invalid. We should not allow that.

OttoWinter (approved):

PR #559, opened 2019-03-22
Fix light flash

placidorevilla:
## Description:

Fixes light flash. Right now, requesting a flash with whatever duration, actually changes the light to the target color, but doesn't revert. Furthermore, the light component saves the flash configuration as the new one, so when rebooting, or turning off/on, the light changes to the flash color.

## Checklist:

### DIFF
--- a/src/esphome/light/light_state.cpp
+++ b/src/esphome/light/light_state.cpp
@@ -38,6 +38,9 @@ void LightState::start_flash(const LightColorValues &target, uint32_t length) {
   if (length == 0)
     return;
 
+  if (this->active_effect_ != nullptr)
+    this->active_effect_->stop();
+
   LightColorValues end_colors = this->values_;
   if (this->transformer_ != nullptr)
     end_colors = this->transformer_->get_end_values();
@@ -63,6 +66,8 @@ LightColorValues LightState::get_current_values() {
       this->remote_values_ = this->values_ = this->transformer_->get_end_values();
       this->transformer_ = nullptr;
       this->send_values();
+      if (this->active_effect_ != nullptr)
+        this->active_effect_->start_internal();
     } else {
       this->values_ = this->transformer_->get_values();
     }
@@ -197,13 +202,15 @@ void LightState::current_values_as_cwww(float color_temperature_cw, float color_
   *warm_white = gamma_correct(*warm_white, this->gamma_correct_);
 }
 void LightState::loop() {
-  if (this->active_effect_ != nullptr) {
+  if (this->active_effect_ != nullptr && this->transformer_ == nullptr) {
     this->active_effect_->apply();
   }
 
   if (this->next_write_ || (this->transformer_ != nullptr && this->transformer_->is_continuous())) {
     this->output_->write_state(this);
     this->next_write_ = false;
+  } else if (this->transformer_ != nullptr) {
+    this->get_current_values();
   }
 }
 LightTraits LightState::get_traits() { return this->output_->get_traits(); }
@@ -468,16 +475,19 @@ void LightState::StateCall::perform() const {
     this->state_->start_effect(*this->effect_);
   }
 
-  LightStateRTCState saved;
-  saved.state = v.get_state() != 0.0f;
-  saved.brightness = v.get_brightness();
-  saved.red = v.get_red();
-  saved.green = v.get_green();
-  saved.blue = v.get_blue();
-  saved.white = v.get_white();
-  saved.color_temp = v.get_color_temperature();
-  saved.effect = *this->state_->active_effect_index_;
-  this->state_->rtc_.save(&saved);
+  // Don't save the new state to preferences if it's just a flash, since it's transient
+  if (!this->flash_length_.has_value()) {
+    LightStateRTCState saved;
+    saved.state = v.get_state() != 0.0f;
+    saved.brightness = v.get_brightness();
+    saved.red = v.get_red();
+    saved.green = v.get_green();
+    saved.blue = v.get_blue();
+    saved.white = v.get_white();
+    saved.color_temp = v.get_color_temperature();
+    saved.effect = *this->state_->active_effect_index_;
+    this->state_->rtc_.save(&saved);
+  }
   this->state_->send_values();
 }
 LightState::StateCall::StateCall(LightState *state) : state_(state) {}


src/esphome/light/light_transformer.cpp
```diff
@@ -59,7 +59,7 @@ LightColorValues LightFlashTransformer::get_end_values() { return this->get_star
 LightFlashTransformer::LightFlashTransformer(uint32_t start_time, uint32_t length, const LightColorValues &start_values,
                                              const LightColorValues &target_values)
     : LightTransformer(start_time, length, start_values, target_values) {}
-bool LightFlashTransformer::is_continuous() { return false; }
+bool LightFlashTransformer::is_continuous() { return true; }
```
OttoWinter: Flashes are not continuous transformers 

> is_continuous(): Whether the output needs to be written in every loop cycle.

If this is to fix some behavior with flash it should be done in the state.

src/esphome/light/light_state.cpp
```diff
@@ -197,7 +202,7 @@ void LightState::current_values_as_cwww(float color_temperature_cw, float color_
   *warm_white = gamma_correct(*warm_white, this->gamma_correct_);
 }
 void LightState::loop() {
-  if (this->active_effect_ != nullptr) {
+  if (this->active_effect_ != nullptr && this->transformer_ == nullptr) {
```
OttoWinter: Please revert this, effects should still be applied when a transformer is active (effects can call transformers)

src/esphome/light/light_state.cpp
```diff
@@ -63,6 +66,8 @@ LightColorValues LightState::get_current_values() {
       this->remote_values_ = this->values_ = this->transformer_->get_end_values();
       this->transformer_ = nullptr;
       this->send_values();
+	  if (this->active_effect_ != nullptr)
+		  this->active_effect_->start_internal();
```
OttoWinter: Revert this, effects are started in `start_effect`

src/esphome/light/light_state.cpp
```diff
@@ -38,6 +38,9 @@ void LightState::start_flash(const LightColorValues &target, uint32_t length) {
   if (length == 0)
     return;
 
+  if (this->active_effect_ != nullptr)
+	  this->active_effect_->stop();
```
OttoWinter: Revert this, effects can call flashes.

src/esphome/light/light_state.cpp
placidorevilla: Then it's impossible to flash while an effect is currently running.

src/esphome/light/light_state.cpp
placidorevilla: This is also to allow flashes while an effect is running.

src/esphome/light/light_state.cpp
placidorevilla: Ditto

src/esphome/light/light_state.cpp
OttoWinter: It is, see random light effect as an example. Light effects do not always write to the state. As seen in the random light effect, they can use the `.turn_on(); ... .perform();` api to call transitions etc.

src/esphome/light/light_transformer.cpp
placidorevilla: From what I can see in the code, if is_continuous is false, the flash is applied in the first iteration of loop, but then the loop ends and it's never reverted back to the previous state. What is your suggestion of how this should be? (is_continuous is very confusing as a property, since you could argue that transitioning from brightness 50% to brightness 50% over 10 seconds is continuous, but then so is a flash, since it has a duration...).

OttoWinter: > but then the loop ends and it's never reverted back to the previous state

See line 63, at the end of a transformer the end values are applied

```cpp
this->remote_values_ = this->values_ = this->transformer_->get_end_values();
```

> but then so is a flash, since it has a duration...

Continuous here means that the transformer has a new value in each loop cycle, like it is for a transition. Flashes only need to override the state two times: At the beginning of the flash, and at the end of the flash.

src/esphome/light/light_state.cpp
placidorevilla: Let me give you an example of what I mean. I have an rgb strip, that I can apply an effect to (let's say rainbow). I also have a user defined service like this:

```
 20   services:
 21     - service: flash
 22       then:
 23         - light.turn_on:
 24             id: light
 25             brightness: 100%
 26             red: 100%
 27             green: 100%
 28             blue: 100%
 29             flash_length: 100ms
```

While I'm running the effect. If I call that service asynchronously, the strip won't flash, it will briefly change brightness, but that's all.

src/esphome/light/light_transformer.cpp
placidorevilla: The flash transformer never says is_finished, because it's not updated ever again, since it's not continuous :-)

OttoWinter: Ah ok now I understand, yes that makes sense :)

On second thought, I think we can just remove `is_continuous` - while a transformer is active it should have complete control of the light. So remove `is_continuous` and change line 204 to 
```cpp
if (this->next_write_ || this->transformer_ != nullptr) {
```

src/esphome/light/light_state.cpp
OttoWinter: I see, but I'd say that's the intended behavior. Flashes (like transitions) should only temporarily affect the light strip, after that the previous state should be exactly continued. So for example flashing red during a rainbow effect should: 
1. Turn strip red
2. Wait for flash time
3. Go back to rainbow effect

If the user wants to cancel the effect they must a) turn off the light or b) call with `effect: "None"`.

Removing `is_continuous` would solve that I think

placidorevilla:
> I see, but I'd say that's the intended behavior. Flashes (like transitions) should only temporarily affect the light strip, after that the previous state should be exactly continued. So for example flashing red during a rainbow effect should:
> 
> Turn strip red
> Wait for flash time
> Go back to rainbow effect
> If the user wants to cancel the effect they must a) turn off the light or b) call with effect: "None".

The current code in this PR implements exactly this behavior. The stop/start of the effect is necessary, since for addressable lights, upon enabling an effect, the AddressableLight object takes full control of the light (via `set_effect_active()`), so the transform doesn't do anything, even if `active_effect_->apply()` is not called.

> 
> Removing is_continuous would solve that I think
> 
> On second thought, I think we can just remove is_continuous - while a transformer is active it should have complete control of the light

Removing `is_continuous` would also work, but the only place this method is called is in this loop anyway, so it would be reasonable to completely remove the method from the abstract class completely. As a drawback, for flashes, it would update the output on every loop, which is not necessary. The code in the PR keeps `is_continuous` and uses it to avoid updating the output, and it just checks if the transform is finished in case it's not continuous.

LMKWYT, I'd keep this PR as it is right now, since I believe it implements the intended behavior without any of the drawbacks, but if you prefer to remove `is_continuous` altogether and pay the penalty of updating the output on every loop while the transition is ongoing, I can do that too.

OttoWinter:
> the AddressableLight object takes full control of the light (via set_effect_active()), so the transform doesn't do anything, even if active_effect_->apply() is not called.

Addressable effects are a completely different thing. Yes they take full control of the strip (and in fact they're kind of a hack). But for non-addressable effects my statement still stands: transition/flash are *temporary* transformers. AND effects can call them internally. Calling a transition or flash should *not* kill the effect.

By that reasoning, we should also kill the effect when a transition begins - but some addressable effects for example take the current RGB value of the light as the base of their data (for example addressable flicker)

> As a drawback, for flashes, it would update the output on every loop, which is not necessary. 

Yes that was the original intent of that property. However, there are no outputs that I'm aware of for which this would be any problem. If that were the case, transitions would never work.

>  since I believe it implements the intended behavior without any of the drawbacks

As I said, this PR will change the behavior in that flashes cancel the effect. Flashes are temporary changes to a light's state and should not affect anything

placidorevilla:
> > the AddressableLight object takes full control of the light (via set_effect_active()), so the transform doesn't do anything, even if active_effect_->apply() is not called.
> 
> Addressable effects are a completely different thing. Yes they take full control of the strip (and in fact they're kind of a hack). But for non-addressable effects my statement still stands: transition/flash are _temporary_ transformers. AND effects can call them internally. Calling a transition or flash should _not_ kill the effect.

The code I propose restarts the effect as soon as the flash is done. If this is not a valid approach, then a flash can't take over the effect. I was maybe thinking of doing something specific for addressable effects (like pausing the effect somehow, but the APIs are protected for that). Honestly, I don't know how to implement the behavior without stopping/restarting the effect.

> 
> By that reasoning, we should also kill the effect when a transition begins - but some addressable effects for example take the current RGB value of the light as the base of their data (for example addressable flicker)

Transitions don't work in this case anyway. If flicker is running and I request a color change, the color is immediately changed (still flickering, though).

> 
> > As a drawback, for flashes, it would update the output on every loop, which is not necessary.
> 
> Yes that was the original intent of that property. However, there are no outputs that I'm aware of for which this would be any problem. If that were the case, transitions would never work.

I'm not saying it would be a problem, but it feels weird to update the output with the same color on every loop for the duration of the flash. If you think this is OK, I can completely remove the is_continuous methods.

> 
> > since I believe it implements the intended behavior without any of the drawbacks
> 
> As I said, this PR will change the behavior in that flashes cancel the effect. Flashes are temporary changes to a light's state and should not affect anything

As I said, the effect restarts immediately after (in fact, it doesn't clear the active_effect_ member, so it can maintain the state, for example, for the scan effect, if flash is called, when it's done, the scan continues from exactly the same pixel).

OttoWinter:
> If this is not a valid approach, then a flash can't take over the effect.

That's the thing, the issue here are addressable effects - *they* do not work like other effects in that they allow transformers to work while they're present. With "normal" effects flashes *do* take over the effect.

> Transitions don't work in this case anyway. If flicker is running and I request a color change, the color is immediately changed (still flickering, though).

Yes, they do not *transition* to the target color, but they do use it as an argument - one of the use cases here is that you can give an addressable effect a color as a *parameter*.

> As I said, the effect restarts immediately after (in fact, it doesn't clear the active_effect_ member, so it can maintain the state, for example, for the scan effect, if flash is called, when it's done, the scan continues from exactly the same pixel).

Oh I did not see that. However, I do not like this approach at all as there could be tons of other edge cases, for example: what happens when a light effect is cancelled while a flash is in effect? stop will be called twice - it may work now but custom effect writers will not except that. I'm sure there are other edge cases too - especially for the light component I like things to be 100% clean with no workarounds because the logic is already very complex.

PR #580, opened 2019-04-03
Add SM16716 output component.

kll:
## Description:
This adds an output component for the SM16716 LED driver. I started with the existing MY9231 output component as a base and adjusted as needed. Hopefully that means the code is already pretty close to acceptable. :)

This chip is used in these bulbs (that I'm aware of):
* Feit Electric A19 Smart WiFi Bulb
* Merkury A21 75W color bulb

I only have the Feit Electric A19 to test with and it only as a single chip so I can't test using multiple chips for additional channels. 

I also haven't made the documentation changes yet. I'll try to get to those soon.

**Related issue (if applicable):** implements esphome/feature-requests#83

**Pull request in [esphome](https://github.com/esphome/esphome) with python changes (if applicable):** esphome/esphome#500
**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#217

## Checklist:

  - [x] Documentation added/updated in [esphome-docs](https://github.com/esphome/esphome-docs).

### DIFF
--- a/src/esphome/defines.h
+++ b/src/esphome/defines.h
@@ -129,6 +129,7 @@
 #define USE_ULN2003
 #define USE_TOTAL_DAILY_ENERGY_SENSOR
 #define USE_MY9231_OUTPUT
+#define USE_SM16716_OUTPUT
 #define USE_CUSTOM_SENSOR
 #define USE_CUSTOM_BINARY_SENSOR
 #define USE_CUSTOM_OUTPUT
@@ -219,6 +220,11 @@
 #endif
 #endif
 #ifdef USE_MY9231_OUTPUT
+#ifndef USE_OUTPUT
+#define USE_OUTPUT
+#endif
+#endif
+#ifdef USE_SM16716_OUTPUT
 #ifndef USE_OUTPUT
 #define USE_OUTPUT
 #endif
--- a/src/esphome/defines.h
+++ b/src/esphome/defines.h
@@ -224,11 +224,6 @@
 #define USE_OUTPUT
 #endif
 #endif
-#ifdef USE_SM16716_OUTPUT
-#ifndef USE_OUTPUT
-#define USE_OUTPUT
-#endif
-#endif
 
 #ifdef USE_APDS9960
 #ifndef USE_SENSOR
--- a/src/esphome/output/sm16716_output_component.cpp
+++ b/src/esphome/output/sm16716_output_component.cpp
@@ -22,11 +22,7 @@ static const char *TAG = "output.sm16716";
 
 SM16716OutputComponent::SM16716OutputComponent(GPIOPin *pin_data, GPIOPin *pin_clock, uint8_t num_channels,
                                                uint8_t num_chips, bool update)
-    : pin_data_(pin_data),
-      pin_clock_(pin_clock),
-      num_channels_(num_channels),
-      num_chips_(num_chips),
-      update_(update) {}
+    : pin_data_(pin_data), pin_clock_(pin_clock), num_channels_(num_channels), num_chips_(num_chips), update_(update) {}
 
 void SM16716OutputComponent::setup() {
   ESP_LOGCONFIG(TAG, "Setting up SM16716OutputComponent...");
@@ -64,13 +60,13 @@ void SM16716OutputComponent::loop() {
     this->write_byte_(this->pwm_amounts_[index]);
     index++;
   }
-  
+
   // send a blank 25 bits to signal the end
   this->write_bit_(false);
   this->write_byte_(0);
   this->write_byte_(0);
   this->write_byte_(0);
-  
+
   this->update_ = false;
 }
 
@@ -119,7 +115,7 @@ SM16716OutputComponent::Channel::Channel(SM16716OutputComponent *parent, uint8_t
     : FloatOutput(), parent_(parent), channel_(channel) {}
 
 void SM16716OutputComponent::Channel::write_state(float state) {
-  uint8_t amount = state * 0xFF;
+  auto amount = uint8_t(state * 0xFF);
   this->parent_->set_channel_value_(this->channel_, amount);
 }
 
--- a/src/esphome/output/sm16716_output_component.cpp
+++ b/src/esphome/output/sm16716_output_component.cpp
@@ -1,0 +1,128 @@
+// Based on:
+//   - The SM16716 implementation from this repository.
+//   - The Tasmota SM16716 implementation: https://github.com/arendst/Sonoff-Tasmota/blob/master/sonoff/xdrv_04_light.ino
+//   - This older implementation: https://github.com/sowbug/sm16716
+//   - The datasheet (in Chinese only): https://github.com/sowbug/sm16716/blob/master/SM16716%20Datasheet%20%5BChinese%5D.pdf
+
+#include "esphome/defines.h"
+
+#ifdef USE_SM16716_OUTPUT
+
+#include "esphome/log.h"
+#include "esphome/output/sm16716_output_component.h"
+
+ESPHOME_NAMESPACE_BEGIN
+
+namespace output {
+
+static const char *TAG = "output.sm16716";
+
+SM16716OutputComponent::SM16716OutputComponent(GPIOPin *pin_mosi, GPIOPin *pin_sclk,
+                                               uint8_t num_channels, uint8_t num_chips,
+                                               bool update)
+    : pin_mosi_(pin_mosi),
+      pin_sclk_(pin_sclk),
+      num_channels_(num_channels),
+      num_chips_(num_chips),
+      update_(update) {}
+
+void SM16716OutputComponent::setup() {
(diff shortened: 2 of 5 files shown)


kll:
That build failure is unrelated to the code. Can you kick it and get it to run again?

EDIT: Nevermind, I'm adding new commits which kicked off the build again.

src/esphome/output/sm16716_output_component.cpp
```diff
@@ -0,0 +1,130 @@
...
+}
+
+void SM16716OutputComponent::write_bit_(bool value) {
+  this->pin_mosi_->digital_write(value);
+  this->pin_sclk_->digital_write(true);
+  this->pin_sclk_->digital_write(false);
+}
+
+void SM16716OutputComponent::write_byte_(uint8_t data) {
+  shiftOut(this->pin_mosi_->get_pin(), this->pin_sclk_->get_pin(), MSBFIRST, data);
```
OttoWinter: Replace this with a direct implementation - this code won't work with inverted pins (if anybody ever needs that) and esphome's pin abstraction should always be used directly.

src/esphome/output/sm16716_output_component.cpp
```diff
@@ -0,0 +1,130 @@
...
+    return;
+  }
+
+  for (uint8_t i = 0; i < 50; i++) {
+    this->write_bit_(false);
+  }
+
+  // send 25 bits (1 start bit plus 24 data bits) for each chip
+  uint8_t index = 0;
+  while (index < this->num_channels_) {
```
OttoWinter: This should be a `for` loop - there's no reason to use while here.

src/esphome/output/sm16716_output_component.cpp
```diff
@@ -0,0 +1,130 @@
...
+  if (this->pwm_amounts_[index] != value) {
+    this->update_ = true;
+  }
+  this->pwm_amounts_[index] = value;
+}
+
+void SM16716OutputComponent::write_bit_(bool value) {
+  this->pin_mosi_->digital_write(value);
+  this->pin_sclk_->digital_write(true);
+  this->pin_sclk_->digital_write(false);
```
OttoWinter: This looks like a very fast loop (prob. up to 10MHz), is there data about how fast the SM16716 can accept data from the bus?

src/esphome/defines.h
```diff
@@ -223,6 +224,11 @@
 #define USE_OUTPUT
 #endif
 #endif
+#ifdef USE_SM16716_OUTPUT
```
OttoWinter: Remove this, it is unnecessary (output will be auto-loaded with this component)

src/esphome/output/sm16716_output_component.cpp
```diff
@@ -0,0 +1,130 @@
...
+  ESP_LOGCONFIG(TAG, "Setting up SM16716OutputComponent...");
+  this->pin_data_->setup();
+  this->pin_data_->digital_write(false);
+  this->pin_clock_->setup();
+  this->pin_clock_->digital_write(false);
+  this->pwm_amounts_.resize(this->num_channels_, 0);
+}
+void SM16716OutputComponent::dump_config() {
+  ESP_LOGCONFIG(TAG, "SM16716:");
+  LOG_PIN("  DATA Pin: ", this->pin_data_);
```
OttoWinter:
```suggestion
  LOG_PIN("  Data Pin: ", this->pin_data_);
```

This shouts too much :)

src/esphome/output/sm16716_output_component.h
```diff
@@ -0,0 +1,91 @@
...
+   *
+   * @param pin_data The data output pin, connected to the SM16716's DIN pin.
+   * @param pin_clock The clock pin, connected to the SM16716's DCLK pin.
+   * @param num_channels Total number of channels of the whole daisy chain.
+   * @param num_chips Number of chips in the daisy chain.
+   * @param update Update/reset duty data at boot (driver will keep
+   *               configuration after powercycle)
+   */
+  SM16716OutputComponent(GPIOPin *pin_data, GPIOPin *pin_clock, uint8_t num_channels = 3, uint8_t num_chips = 1,
+                         bool update = true);
```
OttoWinter: Remove `num_channels`, `num_chips` and `update` from the constructor (they're not used because application.h doesn't define them) and set them up with default arguments in their definition. For example `bool update_{true};`

src/esphome/output/sm16716_output_component.h
```diff
@@ -0,0 +1,91 @@
...
+   * @return The new channel output component.
+   */
+  Channel *create_channel(uint8_t channel, PowerSupplyComponent *power_supply = nullptr, float max_power = 1.0f);
+
+  /// Manually set the total number of channels. Defaults to 3.
+  void set_num_channels(uint8_t num_channels);
+  /// Manually set the number of chips. Defaults to 1.
+  void set_num_chips(uint8_t num_chips);
+  /// Manually set duty data update on boot. Defaults is true.
+  void set_update(bool update);
```
OttoWinter: Do we really need this manual update on boot option? Now that we have restoring from flash on ESP8266 I see little reason for it to exist (just adds more config options and thus makes it harder to use)

src/esphome/output/sm16716_output_component.cpp
```diff
@@ -0,0 +1,126 @@
...
+}
+
+void SM16716OutputComponent::write_bit_(bool value) {
+  this->pin_data_->digital_write(value);
+  this->pin_clock_->digital_write(true);
+  this->pin_clock_->digital_write(false);
+}
+
+void SM16716OutputComponent::write_byte_(uint8_t data) {
+  shiftOut(this->pin_data_->get_pin(), this->pin_clock_->get_pin(), MSBFIRST, data);
```
OttoWinter: Don't use shiftOut here. It won't work when:

- any of the pins are inverted
- esphome moves to a non-arduino base (if it ever does so)

The internal `GPIOPin` should always be used directly if possible.

src/esphome/output/sm16716_output_component.h
```diff
@@ -0,0 +1,91 @@
...
+
+  // ========== INTERNAL METHODS ==========
+  // (In most use cases you won't need these)
+  uint8_t get_num_channels() const;
+  uint8_t get_num_chips() const;
```
OttoWinter: Unless these getters are used anywhere, we should remove them (older components still have getters but those are also slowly being removed)

src/esphome/output/sm16716_output_component.cpp
```diff
@@ -0,0 +1,126 @@
...
+  auto *c = new Channel(this, channel);
+  c->set_power_supply(power_supply);
+  c->set_max_power(max_power);
+  return c;
+}
+
+float SM16716OutputComponent::get_setup_priority() const { return setup_priority::HARDWARE; }
+
+void SM16716OutputComponent::set_channel_value_(uint8_t channel, uint8_t value) {
+  ESP_LOGV(TAG, "set channels %u to %u", channel, value);
```
OttoWinter: This can potentially be called very often, so make it a `ESP_LOGVV` message or remove it.

src/esphome/output/sm16716_output_component.cpp
```diff
@@ -0,0 +1,126 @@
...
+  if (this->pwm_amounts_[index] != value) {
+    this->update_ = true;
+  }
+  this->pwm_amounts_[index] = value;
+}
+
+void SM16716OutputComponent::write_bit_(bool value) {
+  this->pin_data_->digital_write(value);
+  this->pin_clock_->digital_write(true);
+  this->pin_clock_->digital_write(false);
```
OttoWinter, line 94: This loop can run at a very high speed (a few 10s of MHz depending on optimization). Does the datasheet have requirements on timing?

OttoWinter (commented):
Whoops, forgot to press the comment button to publish the review comments. Sorry about that.

src/esphome/output/sm16716_output_component.cpp
kll: Best I can tell by sending the Chinese datasheet through a translator is that the chip can handle 30 MHz. This seems to be the interpretation of the [Tasmota implementation](https://github.com/arendst/Sonoff-Tasmota/blob/master/sonoff/xdrv_04_light.ino#L382) as well. Seems fine at least with the relatively slow 8266 in the Feit bulbs I'm testing with, but I suppose it could be a problem if someone hooked up a chip to a higher clocked ESP32. Is there any good examples of adjusting timing to account for the actual clock speed? I'm pretty new to this level of embedded programming and am mostly in the "copy and tweak" rather than fully original coding level. :)

kll:
Did I miss any of the requested changes?

OttoWinter:
> Did I miss any of the requested changes?

Hi, from a quick look this PR looks quite good. Unfortunately, esphome has had a big refactor recently and all C++ source code moved in the esphome/esphome repository. I have plans to port all existing PRs there at some point, but that will take some time.

PR #583, opened 2019-04-09
Add support for IAQ calculation in BME680

rosterloh:
## Description:
This PR adds an indoor air quality sensor to the BME680 component. I feel this measure is more useful than the raw gas resistance.

## Checklist:

### DIFF
--- a/examples/bme680-test/bme680-test.cpp
+++ b/examples/bme680-test/bme680-test.cpp
@@ -1,0 +1,21 @@
+#include <esphome.h>
+
+using namespace esphome;
+
+void setup() {
+  App.set_name("BME680 Test");
+  App.init_log();
+
+  App.init_wifi("YOUR_SSID", "YOUR_PASSWORD");
+  App.init_mqtt("MQTT_HOST", "USERNAME", "PASSWORD");
+  App.init_ota()->start_safe_mode();
+
+  App.init_i2c(21, 22, true);
+  App.make_bme680_sensor("BME680 Temperature", "BME680 Pressure", "BME680 Humidity", "BME680 Gas Resistance", "BME680 Air Quality");
+  
+  App.setup();
+}
+
+void loop() {
+  App.loop();
+}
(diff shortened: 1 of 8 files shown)


HFsi:
What gas does it measure?

rosterloh:
"Gases that can be detected by the BME680 include Volatile Organic Compounds (VOC) from paints (such as formaldehyde), lacquers, paint strippers, cleaning supplies, furnishings, office equipment, glues, adhesives and alcohol." 

According to the Bosch website

OttoWinter (commented):
This sounds more like something for a lambda sensor to me - see bme280 environment cookbook entry as an example.

Reason is: For all I know this calculation could be entirely arbitrary - components should represent real physical values themselves. All else should be done by external stuff like lambda sensors. 

Another reason is that this appears to depend on when the ESP is powered on, for example if you switch it on while there's more pollution in the air you will get very different values due to the new baseline. In lambda sensors you can at least set a constant for this value so it won't depend on the turn-on-time.

rosterloh:
It's up to you really. This is the generally accepted method for acquiring this measurement and the reason for the baseline is due to allowing for the manufacturer recommended 5 minutes for the internal heater to settle.

OttoWinter:
I understand, however I don't think such a measurement can really have any meaning if it's not stable/reproducible (inner scientist speaking here). I would prefer this to be a template sensor cookbook entry.

HFsi:
@OttoWinter Have you tried the BME680 yourself?
I've found quite difficult gas measurement

rosterloh:
As you wish. Closing.

OttoWinter:
@HFSi yes I have, and I know it doesn't produce any user-usable values directly. However, the inner scientist in me says that this calculation just tries to "look" like a real phyiscally-based value, but in reality it could mean anything.

I haven't tried this formula myself I must say, maybe it does produce good estimations of air quality (though that term would first need to be defined adequately). If someone can point me to research showing this formula is good for approximating air quality, then I'd be for adding this to ESPHome.

Idk, maybe this is also too harsh, but at least I don't see how this formula could somehow magically estimate the air quality as a percentage with only those sensor inputs.

PR #590, opened 2019-04-24
CCS811

huhuhugo1:
support for sensor CCS811. CCS811 is a sensor with switch to set a baseline, timeout support for warming up sensor, status text column and eCO2 and TVOC measuring


OttoWinter (commented):
Hi! Thanks for this new integration!

Unfortunately, ESPHome's codebase has undergone some pretty major changes recently, and this PR will therefore need to be updated. Basically, all esphome source code is now directly hosted in the esphome repo (and this repo here is no longer used). Plus a few minor changes were made regarding the best practices of platforms.

The dev docs are still being written, so you'd have to check the other integrations.
I'd recommend looking at these two integrations:
- https://github.com/esphome/esphome/tree/dev/esphome/components/mhz19 (for multi-sensor components)
- https://github.com/esphome/esphome/tree/dev/esphome/components/tsl2561 (for i2c devices)

I have planned to go through some of the PRs here and port them to the new structure at some point in the future, however that will take some time.

Thanks!

dougmoscrop:
as an outside observer this is an awesome organizational change, woot!

PR #592, opened 2019-05-05
MH-Z19 sensor: disable auto-calibration and  some odd fixes

imammedo:
## Description:
* Adds detection of MH-Z19B sensor
* Ignore sensor reported values until it's booted.
* Disable auto-calibration as it makes reported CO2 values not reflecting real situation in not sufficiently ventilated environment.
* Fix/workaround non-recoverable "Invalid preamble from MHZ19!" case, when sensor is connected to hardware UART.

## Checklist:

  - [ x] The code change follows the [standards](https://esphome.io/guides/contributing.html#contributing-to-esphome-core)

### DIFF
--- a/src/esphome/sensor/mhz19_component.cpp
+++ b/src/esphome/sensor/mhz19_component.cpp
@@ -38,6 +38,13 @@ void MHZ19Component::update() {
     return;
   }
 
+  /* Sensor reports U(15000) during boot, ingnore reported CO2 until it boots */
+  uint16_t u = (response[6] << 8) + response[7];
+  if (u == 15000) {
+    ESP_LOGD(TAG, "Sensor is booting");
+    return;
+  }
+
   uint8_t checksum = mhz19_checksum(response);
   if (response[8] != checksum) {
     ESP_LOGW(TAG, "MHZ19 Checksum doesn't match: 0x%02X!=0x%02X", response[8], checksum);
--- a/src/esphome/sensor/mhz19_component.cpp
+++ b/src/esphome/sensor/mhz19_component.cpp
@@ -27,6 +27,7 @@ uint8_t mhz19_checksum(const uint8_t *command) {
 
 void MHZ19Component::update() {
   uint8_t response[MHZ19_RESPONSE_LENGTH];
+repeat:
   if (!this->mhz19_write_command_(MHZ19_COMMAND_GET_PPM, response)) {
     ESP_LOGW(TAG, "Reading data from MHZ19 failed!");
     this->status_set_warning();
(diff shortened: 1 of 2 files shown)


src/esphome/sensor/mhz19_component.cpp
```diff
@@ -26,18 +27,59 @@ uint8_t mhz19_checksum(const uint8_t *command) {
...
+    ESP_LOGW(TAG, "Invalid preamble from MHZ19! [%0x %0x %0x %0x %0x %0x %0x %0x %0x]",
+      response[0], response[1], response[2], response[3], response[4],
+      response[5], response[6], response[7], response[8]);
     this->status_set_warning();
+    if (++this->cleanup_rx_buffer > 2) {
+        /* try to recover and give up on it if several attemprs were no successful until the next update cycle */
+        this->cleanup_rx_buffer = 0;
+        return;
+    }
+    goto repeat;
```
OttoWinter, line 47: `goto` is a big anti-pattern - could this be changed into a simple for loop?

src/esphome/sensor/mhz19_component.cpp
```diff
@@ -26,18 +27,59 @@ uint8_t mhz19_checksum(const uint8_t *command) {
...
+    if (!this->mhz19_write_command_(MHZ19_COMMAND_ABC_DISABLE, abc_ack)) {
+      ESP_LOGW(TAG, "Failed to read ABC disable ack!");
+      return;
+    }
+    this->abc_disabled = true;
+    /*
+     * TODO: implement an option to recalibrate sensor, to allow for occasional
+     * manual recalibration
+     */
+  }
```
OttoWinter, line 81: This whole block sounds a bit more like it should go in the `setup()` function. We should probably move it there.

src/esphome/sensor/mhz19_component.cpp
```diff
@@ -26,18 +27,59 @@ uint8_t mhz19_checksum(const uint8_t *command) {
...
+    this->model_b = true;
+  }
+
+  if (this->model_b && this->abc_disabled == false) {
+    uint8_t abc_ack[MHZ19_RESPONSE_LENGTH];
+    /*
+     * MH-Z19B allows to enable/disable 'automatic baseline calibration' (datasheet MH-Z19B v1.2),
+     * disable it to prevent sensor baseline drift in not well ventilated area
+     */
+    ESP_LOGI(TAG, "Disabling ABC on boot");
```
OttoWinter, line 71:
```suggestion
    ESP_LOGD(TAG, "Disabling ABC on boot");
```

src/esphome/sensor/mhz19_component.cpp
```diff
@@ -58,6 +100,19 @@ void MHZ19Component::update() {
 
 bool MHZ19Component::mhz19_write_command_(const uint8_t *command, uint8_t *response) {
   this->flush();
+
+  /*
+   * UART0 in NodeMCU v2, sometimes on boot has junk in RX buffer,
+   * check for it and drain all of it before sending commands to sensor
```
OttoWinter, line 106: Uhm the `flush` command should flush the RX buffer (naming is bad, but copied from arduino core). If it does not do so that would be a bug.

OttoWinter (commented):
Thanks! Automatic baseline correction should probably indeed be turned off.

Also See https://github.com/esphome/esphome-core/pull/591#issuecomment-488358598

src/esphome/sensor/mhz19_component.cpp
imammedo: sure  it should be possible

src/esphome/sensor/mhz19_component.cpp
imammedo: I've tried it first but then node hangs for good (had to restart mcu several times before I was able to flash it again) on boot. With UART0  occupied by sensor without hardware to read log from UART1 output, I don't have means to debug it.

I'd still like to move it to setup, but don't know how to fix boot issue.

src/esphome/sensor/mhz19_component.cpp
imammedo: sure, will do n nex pull req

src/esphome/sensor/mhz19_component.cpp
imammedo: That's make on it as well,
but got arduino/esp only a couple days ago so I'm not sure where to look for respective code
and this workaround fixed non working sensor for me.

If it's within this git repo and you can point out to related code, I'll give it a try.

